PR #29, opened 2017-10-06
Version 2.1.0

arun251:
* `LDConfig.Builder#setUseReport` method to allow switching the request verb from `GET` to `REPORT`.
* `LDClient.init` validates that its arguments are non-null.
* Stream connections are closed completely when the app enters background mode.
* Fewer HTTP requests are made to the LaunchDarkly service when feature flags are updated frequently.
* Potential `NullPointerException` in the `variation` methods.
* Removed spurious error when `LDClient` is initialized while the device is offline.


launchdarkly-android-client/src/main/java/com/launchdarkly/android/LDClient.java
```diff
@@ -60,7 +64,21 @@
...
      * @return a {@link Future} which will complete once the client has been initialized.
      */
-    public static synchronized Future<LDClient> init(Application application, LDConfig config, LDUser user) {
+    public static synchronized Future<LDClient> init(@NonNull Application application, @NonNull LDConfig config, @NonNull LDUser user) {
+
+        boolean applicationValid = validateParameter(application);
+        boolean configValid = validateParameter(config);
+        boolean userValid = validateParameter(user);
+        if (!applicationValid) {
+            return Futures.immediateFailedFuture(new LaunchDarklyException("Client initialization requires a valid application"));
```
wokkaflokka: 💯

PR #31, opened 2017-11-21
Explicitly creating the handler with Looper.mainLooper() so the clien…

jonathanmgrimm:
…t can be initialized not on the main thread


arun251:
This looks great @jonathanmgrimm . We will release this shortly.

arun251:
Released in version 2.1.1.

jonathanmgrimm:
thanks!

PR #61, opened 2018-11-28
Added support for TLS v1.2 on pre-Lollipop devices

gefilte:
Android devices _subsequent to_ *Ice Cream Sandwich* (4.0/API level 14) and _prior to_ *Lollipop* (5.0/API level 21) ship with TLS v1.2 protocol and cipher support available, but it is not enabled by default. This change will enable this protocol for these devices, so they can continue to use LaunchDarkly even if the service-side requires TLS v1.2.

### DIFF
--- a/launchdarkly-android-client/src/main/java/com/launchdarkly/android/EventProcessor.java
+++ b/launchdarkly-android-client/src/main/java/com/launchdarkly/android/EventProcessor.java
@@ -49,7 +49,7 @@ class EventProcessor implements Closeable {
         this.consumer = new Consumer(config);
         this.summaryEventSharedPreferences = summaryEventSharedPreferences;
 
-        client = new OkHttpClient.Builder()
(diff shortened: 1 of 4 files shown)


gefilte:
Can we get this PR some attention?

torchhound:
Hi @gefilte sorry for the delay. We internally developed support for TLS v1.2 on pre-Lollipop devices and that feature was just released in version 2.6.0. Thank you for using and supporting LaunchDarkly!

gefilte:
Great, I see your commits – when are you going to cut/publish the release?

torchhound:
We plan on having 2.6.0 available as a dependency within the next several days.

gwhelanLD:
V2.6.0 is now released, and should be available as a gradle dependency currently. Thanks again for your work on this PR @gefilte.

gefilte:
Team resolved similarly.

Issue #83, opened 2019-07-31
Fatal Exception: java.lang.SecurityException on Samsung devices


paulfthomas:
**Describe the bug**
Crash only impacting Samsung devices, probably related to the way the `PendingIntent` is retrieved in the `PollingUpdater`. See https://github.com/launchdarkly/android-client-sdk/blob/master/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/android/PollingUpdater.java#L56

**Logs**
```
Fatal Exception: java.lang.SecurityException: !@Too many alarms (500) registered from uid 10085
       at android.os.Parcel.createException + 1966(Parcel.java:1966)
       at android.os.Parcel.readException + 1934(Parcel.java:1934)
       at android.os.Parcel.readException + 1884(Parcel.java:1884)
       at android.app.IAlarmManager$Stub$Proxy.set + 240(IAlarmManager.java:240)
       at android.app.AlarmManager.setImpl + 722(AlarmManager.java:722)
       at android.app.AlarmManager.setInexactRepeating + 823(AlarmManager.java:823)
       at com.launchdarkly.android.PollingUpdater.startPolling + 36(PollingUpdater.java:36)
       at com.launchdarkly.android.PollingUpdater.startBackgroundPolling + 27(PollingUpdater.java:27)
       at com.launchdarkly.android.ConnectivityManager.startBackgroundPolling + 182(ConnectivityManager.java:182)
       at com.launchdarkly.android.ConnectivityManager.attemptTransition + 258(ConnectivityManager.java:258)
       at com.launchdarkly.android.ConnectivityManager.access$300 + 18(ConnectivityManager.java:18)
       at com.launchdarkly.android.ConnectivityManager$1.run + 64(ConnectivityManager.java:64)
       at com.launchdarkly.android.Throttler$1.run + 49(Throttler.java:49)
       at android.os.Handler.handleCallback + 873(Handler.java:873)
       at android.os.Handler.dispatchMessage + 99(Handler.java:99)
       at android.os.Looper.loop + 214(Looper.java:214)
       at android.os.HandlerThread.run + 65(HandlerThread.java:65)

Caused by android.os.RemoteException: Remote stack trace:
	at com.android.server.SamsungAlarmManagerService.checkMaliciousAppLocked(SamsungAlarmManagerService.java:306)
	at com.android.server.AlarmManagerService.setImpl(AlarmManagerService.java:1748)
[...]

PR #87, opened 2019-10-17
Use FLAG_UPDATE_CURRENT to create PendingIntent

sambroad:
Potential fix for: https://github.com/launchdarkly/android-client-sdk/issues/83
based on https://stackoverflow.com/a/29703161

### DIFF
--- a/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/android/PollingUpdater.java
+++ b/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/android/PollingUpdater.java
@@ -53,7 +53,7 @@ public class PollingUpdater extends BroadcastReceiver {
     }
 
     private static PendingIntent getPendingIntent(Context context) {
-        return PendingIntent.getBroadcast(context, 0, getAlarmIntent(context), 0);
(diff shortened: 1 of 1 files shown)


gwhelanLD:
Hi @sambroad,

Thanks for this PR. This is definitely an approach we are looking at for fixing #83, but we need to do a bit more investigation to validate that this is the approach we want to take on this issue. 

We have not prioritized this work yet as initial triaging has indicated the conditions for triggering this issue should limit the impact substantially. That said, we hope to take on this work soon.

Thanks,
@gwhelanLD

sambroad:
Thanks for your answer. I appreciate you looking into this, and understand you need to be cautious not to introduce any side effects. 
This fix solved the same issue when we applied it in our codebase.
https://github.com/launchdarkly/android-client-sdk/issues/83 is now the most reported crash in our app, so I look forward to seeing it prioritized soon.

gwhelanLD:
We've just released version 2.9.1 of the Android SDK, which should address #83, although in a different manner than proposed in this PR. Please see the discussion on the issue for more information on our investigation and our ultimate solution.

I'm going to close this PR as the issue it was intended to address has now been closed. Thanks again for suggesting this change, even though we ultimately went with a different approach.

PR #201, opened 2022-12-21
prepare 3.3.1 release

LaunchDarklyReleaseBot:
### Fixed:
- If the application is in the background when the SDK is started, the SDK will go into polling mode and immediately make a flag data request to LaunchDarkly. Previously, in this scenario the first poll would not happen until the background poll interval elapsed, so the SDK would effectively never have flag data at initialization time for an app or service that started in the background.


louis-launchdarkly (changes requested):
The change should be good, there is one contract test that is failing, looks like a one-off failure?

louis-launchdarkly (approved):

PR #260, opened 2024-03-15
ci: enabling ci tests after fixing test race condition

tanderson-ld:
**Related issues**

236700

Eliminated recording of evaluation GET command which was leading to race condition in unit tests.


launchdarkly-android-client-sdk/src/androidTest/java/com/launchdarkly/sdk/android/LDClientEventTest.java
```diff
@@ -257,21 +257,16 @@ public void testEventContainsAutoEnvAttributesWhenEnabled() throws Exception {
                     .diagnosticOptOut(true)
                     .serviceEndpoints(Components.serviceEndpoints()
                             .events(baseUrl.uri())
-                            .streaming(baseUrl.uri())
-                            .polling(baseUrl.uri())
```
tanderson-ld, line 261: This was the root cause of the issue.  It was resulting in flag evaluation requests getting sent to the same place as events during the test.  This results in a race condition between checking the event request.

keelerm84 (approved):

PR #261, opened 2024-03-22
fix: improved LDClient.identify(...) behavior when offline

tanderson-ld:
**Requirements**

- [x] I have validated my changes against all supported platform versions

**Related issues**

SC-237349

Updated code so that ContextDataManager switches context independently of ConnectivityManager and ConnectivityManager is not in the path of context switching.

### DIFF
--- a/launchdarkly-android-client-sdk/src/androidTest/java/com/launchdarkly/sdk/android/LDClientEndToEndTest.java
+++ b/launchdarkly-android-client-sdk/src/androidTest/java/com/launchdarkly/sdk/android/LDClientEndToEndTest.java
@@ -144,4 +144,38 @@ public class LDClientEndToEndTest {
             assertEquals(flagValue, client.stringVariation(flagKey, "default"));
         }
     }
+
+    @Test
+    public void identifyWhenPollingFailsAndCacheAlreadyExists() throws Exception {
+        // set up data store with flag data for ContextA and ContextB
+        // insert ContextA's flags
+        LDContext contextA = LDContext.create("ContextA");
+        String flagKeyA = "flag-keyA", flagValueA = "stored-valueA";
+        Flag flagA = new FlagBuilder(flagKeyA).version(1).value(LDValue.of(flagValueA)).build();
+        TestUtil.writeFlagUpdateToStore(store, MOBILE_KEY, contextA, flagA);
+        // insert contextB's flags
+        LDContext contextB = LDContext.create("ContextB");
+        String flagKeyB = "flag-keyB", flagValueB = "stored-valueB";
+        Flag flagB = new FlagBuilder(flagKeyB).version(1).value(LDValue.of(flagValueB)).build();
+        TestUtil.writeFlagUpdateToStore(store, MOBILE_KEY, contextB, flagB);
+
+        // response to initialization for ContextA
+        mockPollingServer.enqueue(new MockResponse().setResponseCode(401));
+
+        // response to contextB's identify
+        mockPollingServer.enqueue(new MockResponse().setResponseCode(401));
+
+        LDConfig config = baseConfig()
+                .dataSource(Components.pollingDataSource())
+                .serviceEndpoints(Components.serviceEndpoints().polling(mockPollingServerUri))
+                .build();
+
+        LDClient client = LDClient.init(application, config, contextA, 1);
+        assertFalse("client should not have been initialized", client.isInitialized());
+        assertFalse("client was offline", client.isOffline());
+        assertEquals(flagValueA, client.stringVariation(flagKeyA, "defaultA"));
+
+        client.identify(contextB).get();
+        assertEquals(flagValueB, client.stringVariation(flagKeyB, "defaultB"));
+    }
 }
--- a/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/AndroidPlatformState.java
+++ b/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/AndroidPlatformState.java
@@ -160,7 +160,12 @@ final class AndroidPlatformState implements PlatformState {
     public void close() {
         connectivityChangeListeners.clear();
         foregroundChangeListeners.clear();
-        application.unregisterReceiver(connectivityReceiver);
+        try {
+            application.unregisterReceiver(connectivityReceiver);
+        } catch (IllegalArgumentException e) {
+            // getting here just means the receiver wasn't registered, which is our objective
+        }
+
         application.unregisterActivityLifecycleCallbacks(lifecycleCallbacks);
     }
 
--- a/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ClientContextImpl.java
+++ b/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ClientContextImpl.java
@@ -123,6 +123,21 @@ final class ClientContextImpl extends ClientContext {
         );
     }
 
+    /**
+     * Sets the evaluation context and returns a new instance of {@link ClientContextImpl}
+     * @param context to now use as the evaluation context
+     * @return a new instance
+     */
+    public ClientContextImpl setEvaluationContext(LDContext context) {
+        return new ClientContextImpl(
+            super.setEvaluationContext(context),
+            this.diagnosticStore,
+            this.fetcher,
+            this.platformState,
+            this.taskExecutor
+        );
+    }
+
     public DiagnosticStore getDiagnosticStore() {
         return diagnosticStore;
     }
--- a/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ConnectivityManager.java
+++ b/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ConnectivityManager.java
@@ -54,7 +54,6 @@ class ConnectivityManager {
     private final DataSourceUpdateSink dataSourceUpdateSink;
     private final ConnectionInformationState connectionInformation;
     private final PersistentDataStoreWrapper.PerEnvironmentData environmentStore;
-    private final ContextDataManager contextDataManager;
     private final EventProcessor eventProcessor;
     private final PlatformState.ForegroundChangeListener foregroundListener;
     private final PlatformState.ConnectivityChangeListener connectivityChangeListener;
@@ -66,7 +65,7 @@ class ConnectivityManager {
     private final AtomicBoolean started = new AtomicBoolean();
     private final AtomicBoolean closed = new AtomicBoolean();
     private final AtomicReference<DataSource> currentDataSource = new AtomicReference<>();
-    private final AtomicReference<LDContext> currentEvaluationContext = new AtomicReference<>();
+    private final AtomicReference<LDContext> currentContext = new AtomicReference<>();
     private final AtomicReference<Boolean> previouslyInBackground = new AtomicReference<>();
     private final LDLogger logger;
     private volatile boolean initialized = false;
@@ -76,19 +75,23 @@ class ConnectivityManager {
     // data is stored; 2. to implement additional logic that does not depend on what kind of data
     // source we're using, like "if there was an error, update the ConnectionInformation."
     private class DataSourceUpdateSinkImpl implements DataSourceUpdateSink {
+        private final ContextDataManager contextDataManager;
         private final AtomicReference<ConnectionMode> connectionMode = new AtomicReference<>(null);
         private final AtomicReference<LDFailure> lastFailure = new AtomicReference<>(null);
 
+        DataSourceUpdateSinkImpl(ContextDataManager contextDataManager) {
+            this.contextDataManager = contextDataManager;
+        }
+
         @Override
-        public void init(Map<String, DataModel.Flag> items) {
-            contextDataManager.initData(currentEvaluationContext.get(),
-                    EnvironmentData.usingExistingFlagsMap(items));
+        public void init(LDContext context, Map<String, DataModel.Flag> items) {
+            contextDataManager.initData(context, EnvironmentData.usingExistingFlagsMap(items));
             // Currently, contextDataManager is responsible for firing any necessary flag change events.
         }
 
         @Override
-        public void upsert(DataModel.Flag item) {
-            contextDataManager.upsert(item);
+        public void upsert(LDContext context, DataModel.Flag item) {
+            contextDataManager.upsert(context, item);
             // Currently, contextDataManager is responsible for firing any necessary flag change events.
         }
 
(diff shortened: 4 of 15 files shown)


launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/AndroidPlatformState.java
```diff
@@ -160,7 +160,12 @@ public File getCacheDir() {
     public void close() {
         connectivityChangeListeners.clear();
         foregroundChangeListeners.clear();
-        application.unregisterReceiver(connectivityReceiver);
+        try {
+            application.unregisterReceiver(connectivityReceiver);
+        } catch (IllegalArgumentException e) {
+            // getting here just means the receiver wasn't registered, which is our objective
+        }
```
tanderson-ld, line 167: For reviewers: AndroidPlatformState claims to be closeable, in which close is idempotent after the first call.  This changes makes it idempotent.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ClientContextImpl.java
```diff
@@ -123,6 +123,21 @@ public static ClientContextImpl forDataSource(
...
+     */
+    public ClientContextImpl setEvaluationContext(LDContext context) {
+        return new ClientContextImpl(
+            super.setEvaluationContext(context),
+            this.diagnosticStore,
+            this.fetcher,
+            this.platformState,
+            this.taskExecutor
+        );
+    }
```
tanderson-ld, line 139: For reviewers:  LDClient's ClientContextImpl member variable should be the authority on the current context.  This allows us to change the context in that object so that can be true after an identify call.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ConnectivityManager.java
```diff
@@ -76,19 +75,23 @@ class ConnectivityManager {
     // data is stored; 2. to implement additional logic that does not depend on what kind of data
     // source we're using, like "if there was an error, update the ConnectionInformation."
     private class DataSourceUpdateSinkImpl implements DataSourceUpdateSink {
+        private final ContextDataManager contextDataManager;
```
tanderson-ld, line 78: For reviewers:  The sink now holds the reference to the data manager instead of a private class holding a reference to an enclosing class's member.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ConnectivityManager.java
```diff
@@ -76,19 +75,23 @@ class ConnectivityManager {
...
+        DataSourceUpdateSinkImpl(ContextDataManager contextDataManager) {
+            this.contextDataManager = contextDataManager;
+        }
+
         @Override
-        public void init(Map<String, DataModel.Flag> items) {
-            contextDataManager.initData(currentEvaluationContext.get(),
-                    EnvironmentData.usingExistingFlagsMap(items));
+        public void init(LDContext context, Map<String, DataModel.Flag> items) {
+            contextDataManager.initData(context, EnvironmentData.usingExistingFlagsMap(items));
```
tanderson-ld, line 88: For reviewers:  The context was added to go along with the flag data.  This helps other parts of the system make decisions on whether or not to accept the flag data.  Ideally, this manager and the sources of this data could only exist when getting data from them makes sense, but a refactoring of the managers to be instead immutable instances and implement disposable is too risky at this time.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ConnectivityManager.java
```diff
@@ -357,13 +367,6 @@ synchronized boolean startUp(@NonNull Callback<Void> onCompletion) {
             return false;
         }
         initialized = false;
-
-        // Calling initFromStoredData updates the current flag state *if* stored flags exist for
-        // this context. If they don't, it has no effect. Currently we do *not* return early from
-        // initialization just because stored flags exist; we're just making them available in case
-        // initialization times out or otherwise fails.
-        contextDataManager.initFromStoredData(currentEvaluationContext.get());
```
tanderson-ld, line 365: For reviewers:  The ConnectivityManager should not be involved at all with the concept of using cached data.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ContextDataManager.java
```diff
@@ -162,21 +139,21 @@ private void initDataInternal(
...
-            logger.debug("Updated flag data for context {} in persistent store", contextId);
+            for (String removedContextId: removedContextIds) {
+                environmentStore.removeContextData(removedContextId);
+                logger.debug("Removed flag data for context {} from persistent store", removedContextId);
+            }
+            if (writeFlagsToPersistentStore && maxCachedContexts != 0) {
+                environmentStore.setContextData(contextId, newData);
+                logger.debug("Updated flag data for context {} in persistent store", contextId);
+            }
+            environmentStore.setIndex(newIndex);
```
tanderson-ld, line 151: For reviewers:  `environmentStore.setIndex(newIndex);` should be in the lock block.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ContextDataManager.java
```diff
@@ -162,21 +139,21 @@ private void initDataInternal(
             newIndex = index.updateTimestamp(contextId, System.currentTimeMillis())
                     .prune(maxCachedContexts, removedContextIds);
             index = newIndex;
-            flagsContextId = contextId;
```
tanderson-ld, line 165: For reviewers:  Calculating a context id is fast and trivial.  Why risk it being outdated?  Removed.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ContextDataManager.java
```diff
@@ -122,26 +115,6 @@ public void initData(
...
-    /**
-     * Attempts to initialize the flag data state from the persistent store. If there was a set of
-     * stored flag data for this context, it updates the current flag state and returns true; also,
-     * the context is added to the list of stored contexts-- evicting old context data if necessary.
-     * If there was no stored data, it leaves the current flag state as is and returns false.
-     *
-     * @param context the new context
-     * @return true if successful
-     */
-    public boolean initFromStoredData(@NonNull LDContext context) {
```
tanderson-ld, line 134: For reviewers:  This functionality was moved to switchToContext.  This also helps to privatize the concept of stored data so that other components don't have to be involved.  Yay separation of concerns.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ContextDataManager.java
```diff
@@ -49,62 +49,55 @@ final class ContextDataManager {
...
         this.environmentStore = environmentStore;
         this.maxCachedContexts = maxCachedContexts;
         this.taskExecutor = ClientContextImpl.get(clientContext).getTaskExecutor();
         this.logger = clientContext.getBaseLogger();
+        switchToContext(clientContext.getEvaluationContext());
     }
 
     /**
-     * Returns the current context.
+     * Switches to providing flag data for the provided context.
```
tanderson-ld, line 75: For reviewers: You can no longer switch context without switching the flags this DataManager will use.  Weird that was ever an option to begin with...

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ContextDataManager.java
```diff
@@ -152,8 +125,12 @@ private void initDataInternal(
         EnvironmentData oldData;
         ContextIndex newIndex;
 
-        synchronized (writerLock) {
-            currentContext = context;
+        synchronized (lock) {
+            if (!context.equals(currentContext)) {
+                // if incoming new data is not for the current context, reject it.
+                return;
+            }
```
tanderson-ld, line 132: For reviewers:  I considered refactoring this code to be willing to store the data for the non-current context if such a situation was to occur, but it seemed unlikely and the refactor would be without risk.  There is also an interesting edge case / race condition if that non-current context was not already cached and it put the cache over its cacheLimit.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/LDClient.java
```diff
@@ -402,8 +400,15 @@ public Future<Void> identify(LDContext context) {
...
                                   Callback<Void> onCompleteListener) {
-        contextDataManager.setCurrentContext(context);
-        connectivityManager.setEvaluationContext(context, onCompleteListener);
+
+        clientContextImpl = clientContextImpl.setEvaluationContext(context);
+
+        // Calling initFromStoredData updates the current flag state *if* stored flags exist for
+        // this context. If they don't, it has no effect. Currently we do *not* return early from
+        // initialization just because stored flags exist; we're just making them available in case
+        // initialization times out or otherwise fails.
```
tanderson-ld, line 409: Need to remove copy-pasta'd comment.

tanderson-ld:
Need to add unit test and end to end integration test.

launchdarkly-android-client-sdk/src/test/java/com/launchdarkly/sdk/android/ContextDataManagerListenersTest.java
```diff
@@ -78,7 +80,7 @@ public void listenerIsNotCalledAfterUnregistering() throws InterruptedException
         manager.registerAllFlagsListener(allFlagsListener);
         manager.unregisterAllFlagsListener(allFlagsListener);
 
-        manager.upsert(flag);
+        manager.upsert(LDContext.builder("aKey").build(), flag);
```
kinyoklion, line 83: I'm not confident in this test.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ClientContextImpl.java
kinyoklion: I have some concerns with this. Copies of the client context are passed to various components, and those client contexts do not have a centralized update mechanism. So putting the evaluation context into the client context provides easy access to a stale or unset LDContext.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/LDClient.java
```diff
@@ -46,17 +46,14 @@ public class LDClient implements LDClientInterface, Closeable {
     // A map of each LDClient (one per environment), or null if `init` hasn't been called yet.
     // Will only be set once, during initialization, and the map is considered immutable.
     static volatile Map<String, LDClient> instances = null;
-    private static volatile PlatformState sharedPlatformState;
-    private static volatile IEnvironmentReporter environmentReporter;
-    private static volatile TaskExecutor sharedTaskExecutor;
-    private static volatile IContextModifier autoEnvContextModifier;
-    private static volatile IContextModifier anonymousKeyContextModifier;
+    private static ClientContextImpl clientContextImpl;
```
kinyoklion, line 49: This needs to be synchronized in some way. Volatile, atomic, etc. Otherwise there is not going to be fencing and you cannot be sure you are starting with the current value or a value that has been cached.

kinyoklion: (Setting a reference itself is atomic, but getting the same reference from a different thread after the reference is set it a different matter.)

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ClientContextImpl.java
tanderson-ld: You are correct.  I believe that is how it has been.  Any components that use the context have an alternative mechanism for updating it when it changes.

launchdarkly-android-client-sdk/src/test/java/com/launchdarkly/sdk/android/ContextDataManagerListenersTest.java
tanderson-ld: Your spidey senses were correct.  Fixed.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/LDClient.java
tanderson-ld: Fixed

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ContextDataManager.java
```diff
@@ -49,62 +49,55 @@ final class ContextDataManager {
...
+        }
 
-     * @param newContext the new context
-     */
-    public void setCurrentContext(@NonNull LDContext newContext) {
-        currentContext = newContext;
-    }
+        EnvironmentData storedData = getStoredData(currentContext);
+        if (storedData == null) {
+            logger.debug("No stored flag data is available for this context");
```
kinyoklion, line 92: Yay logging.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ContextDataManager.java
kinyoklion: In the long-term I think we should attempt to change this to a queue of operations that apply to the store. Maybe transactional in some way. It would be preferable to not have to lock for the duration of IO.

launchdarkly-android-client-sdk/src/androidTest/java/com/launchdarkly/sdk/android/LDClientEndToEndTest.java
```diff
@@ -144,4 +144,38 @@ public void clientUsesStoredFlagsIfInitializationTimesOutInPollingMode() throws
...
+        // response to contextB's identify
+        mockPollingServer.enqueue(new MockResponse().setResponseCode(401));
+
+        LDConfig config = baseConfig()
+                .dataSource(Components.pollingDataSource())
+                .serviceEndpoints(Components.serviceEndpoints().polling(mockPollingServerUri))
+                .build();
+
+        LDClient client = LDClient.init(application, config, contextA, 1);
+        assertFalse("client should not have been initialized", client.isInitialized());
```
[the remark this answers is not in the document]
tanderson-ld: Yes, see line [119](https://github.com/launchdarkly/android-client-sdk/pull/261/files/2b93d4de0600b5f65f2acf1415934e9e730c33c4#diff-f4e86d7e245150ffd9b98a3a63cd85a1a16334f2f4b19f61129e90dcdfdf262dR119) .

kinyoklion (approved):

PR #274, opened 2024-07-09
fix: registered LDStatusListeners are now sent updates that result from a call to LDClient.identify(...)

tanderson-ld:
**Requirements**

- [x] I have validated my changes against all supported platform versions

**Related issues**

https://app.shortcut.com/launchdarkly/story/211876/android-sdk-connectioninformation-and-listeners-not-updated-correctly

The issue was introduced in version 4.X and is has affected 2 major versions.  The root of the issue is that from 3.x to 4.x, the code that sets the connection mode and kicks the last successful connection timestamp was moved out of the connectivity manager, so reconnects (which are handled inside the connectivity manager), were not handled.  The fix is to adjust the code so that cconnection information / connection timestamps are kicked when the reconnect logic succeeds or fails.

Refactoring to match iOS's logic of emitting an offline event when reconnecting.  This would more closely match 3.x.  This was rejected as not having a good ratio of work required and risk involved.  It also could introduce unexpected behavior for customers that implemented LDStatusListeners in version 4.x and 5.x.

### DIFF
--- a/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ConnectivityManager.java
+++ b/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ConnectivityManager.java
@@ -76,8 +76,6 @@ class ConnectivityManager {
     // source we're using, like "if there was an error, update the ConnectionInformation."
     private class DataSourceUpdateSinkImpl implements DataSourceUpdateSink {
         private final ContextDataManager contextDataManager;
-        private final AtomicReference<ConnectionMode> connectionMode = new AtomicReference<>(null);
-        private final AtomicReference<LDFailure> lastFailure = new AtomicReference<>(null);
 
         DataSourceUpdateSinkImpl(ContextDataManager contextDataManager) {
             this.contextDataManager = contextDataManager;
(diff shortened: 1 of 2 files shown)


launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ConnectivityManager.java
```diff
@@ -76,8 +76,6 @@ class ConnectivityManager {
     // source we're using, like "if there was an error, update the ConnectionInformation."
     private class DataSourceUpdateSinkImpl implements DataSourceUpdateSink {
         private final ContextDataManager contextDataManager;
-        private final AtomicReference<ConnectionMode> connectionMode = new AtomicReference<>(null);
-        private final AtomicReference<LDFailure> lastFailure = new AtomicReference<>(null);
```
tanderson-ld, line 80: For reviewers: These pieces of data duplicated what was held in the ConnectionInformationState member of the ConnectivityManager, so removed duplication to reduce chance of being out of sync.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ConnectivityManager.java
```diff
@@ -97,40 +95,10 @@ public void upsert(LDContext context, DataModel.Flag item) {
...
-                }
-                updateStatusListeners(connectionInformation);
-                if (failure != null) {
-                    updateListenersOnFailure(failure);
-                }
+            if (error == null) {
+                updateConnectionInfoForSuccess(newConnectionMode);
+            } else {
+                updateConnectionInfoForError(newConnectionMode, error);
             }
```
tanderson-ld, line 102: For reviewers: Majority of the removed code has been moved into `updateConnectionInfoForSuccess` and `updateConnectionInfoForError`.

louis-launchdarkly (approved):

PR #307, opened 2025-07-21
fix: prevent flag change listener from being invoked with null flag key

tanderson-ld:


launchdarkly-android-client-sdk/src/androidTest/java/com/launchdarkly/sdk/android/LDClientPluginsTest.java
tanderson-ld, line 1: For reviewers: Fixing flaky test due to hash map iteration order not being deterministic across runs.

louis-launchdarkly (approved):
As discussed, while it would be the best to know exactly what is triggering this, if we have a null key at this point, we can't find the correct listener to call anyway and it is better to silence it than crash.

PR #323, opened 2026-02-21
feat: Reduce streaming DNS failures with stale-fallback DNS cache

abelonogov-ld:
## Summary

- Add `CachingDns`, a thread-safe OkHttp `Dns` wrapper that caches successful lookups (10-min TTL) and returns stale cached addresses when a fresh resolution fails — preventing `UnknownHostException` from killing the stream during network transitions
- Share a `ConnectionPool` and the DNS resolver across `StreamingDataSource` restarts (context switches, foreground/background toggles, network changes) via `StreamingDataSourceBuilderImpl`, so cached state survives data source recreation
- Explicitly set `retryOnConnectionFailure(true)` on the streaming OkHttpClient on all API levels

## Background

We observed a high percentage of DNS failures in streaming connections. The root cause is that `StreamingDataSource` creates a new `OkHttpClient` on every `start()` call, and `ConnectivityManager` restarts the data source on every network change — exactly when DNS is most fragile. OkHttp uses `Dns.SYSTEM` (`InetAddress.getAllByName`) with no caching, and Android's system DNS cache has very short TTLs (sometimes ~2 seconds) that get cleared on network transitions.

This pattern of application-level DNS caching with stale fallback is well-established: Alibaba's HTTPDNS SDK, gRPC-Java's `DnsNameResolver`, and Square's own `DnsOverHttps` module all implement similar approaches. Google validated the pattern by adding `DnsOptions.StaleDnsOptions` to the Android framework in API 34.

## Test plan

- [x] Unit tests for `CachingDns`: fresh resolution, cache hits within TTL, TTL expiry refresh, stale fallback on failure, cold failure propagation, per-hostname isolation, expiration boundary
- [x] Existing `StreamingDataSourceTest` passes (builder creates data source with new constructor args transparently)
- [ ] Verify via logs that `CachingDns` warns on stale fallback and that stream reconnects succeed during network changes

---

> [!NOTE]
> **Medium Risk**
> Touches streaming network connection setup and DNS resolution behavior; incorrect caching/pooling could cause connectivity regressions or use stale IPs longer than intended.
> 
> **Overview**
> Adds `CachingDns`, an OkHttp `Dns` wrapper that caches successful lookups with a TTL and falls back to stale cached addresses when fresh resolution fails, reducing `UnknownHostException` disruptions during mobile network transitions.
> 
> Updates streaming to **reuse** a shared DNS resolver and `ConnectionPool` across `StreamingDataSource` restarts, and wires these into the EventSource OkHttp client configuration (including enabling `retryOnConnectionFailure(true)`). Includes unit tests covering cache hit/expiry behavior, stale fallback, per-host caching, and eviction behavior when exceeding `MAX_ENTRIES`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0a6a23e13e6c627d45033ab8dd0f2086b6b1770a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/CachingDns.java
+++ b/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/CachingDns.java
@@ -6,6 +6,8 @@ import com.launchdarkly.logging.LDLogger;
 
 import java.net.InetAddress;
 import java.net.UnknownHostException;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 
@@ -42,7 +44,7 @@ final class CachingDns implements Dns {
         final long expiresAtMs;
 
         CacheEntry(List<InetAddress> addresses, long expiresAtMs) {
-            this.addresses = addresses;
+            this.addresses = Collections.unmodifiableList(new ArrayList<>(addresses));
             this.expiresAtMs = expiresAtMs;
         }
 
--- a/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/CachingDns.java
+++ b/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/CachingDns.java
@@ -93,7 +93,12 @@ final class CachingDns implements Dns {
     }
 
     private void evictExpired(long now) {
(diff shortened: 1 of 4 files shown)


tanderson-ld (commented):
I think this needs more discussion in slack channel so we're on all the same page about possible impact and how we'll confirm there is no negative impact.

PR #339, opened 2026-03-31
chore: Handle Fallback From FDv2 to FDv1

aaron-zeisler:
## The Goal

When the LaunchDarkly backend determines that a client should not be using FDv2, it signals this by including an `x-ld-fd-fallback: true` header in HTTP responses. This PR teaches the Android SDK to detect that header and seamlessly switch from FDv2 data sources to an FDv1 polling fallback, so the SDK continues to receive flag data without interruption.

JIRA ticket: https://launchdarkly.atlassian.net/browse/SDK-1829

## The Approach

The implementation follows a layered approach, working from the bottom of the stack upward:

1. **Data model** — The result types that flow between components gained an `fdv1Fallback` boolean, so any component that receives a server response can carry the fallback signal forward without losing it. This mirrors the js-core approach, where the fallback signal is carried on the data source result type rather than handled as a side channel.

2. **Header detection** — The polling requestor and streaming synchronizer both read the `x-ld-fd-fallback` header and attach it to every result they produce — success, error, or otherwise. This means the signal is never silently dropped. Both js-core and java-core detect the header at the transport layer; the Android implementation follows the same pattern.

3. **Fallback synchronizer** — A new standalone FDv1 polling synchronizer fetches flag data from the legacy `/sdk/evalx` endpoints and produces results in the FDv2 format. It's intentionally decoupled from the existing `PollingDataSource` / `HttpFeatureFlagFetcher` stack because those are tightly coupled to FDv1's `DataSourceUpdateSink`. This matches js-core's approach of implementing a dedicated FDv1 polling data source rather than reusing the existing one.

4. **Orchestration** — The FDv2 data source's run loop checks for the fallback signal after processing each result. When detected, it blocks all FDv2 synchronizers, unblocks the FDv1 slot, and lets the normal synchronizer selection logic pick it up. No special "fallback mode" — it reuses the existing synchronizer cycling mechanism. The `SourceManager.fdv1Fallback()` implementation follows js-core's pattern of resetting the synchronizer index to -1 so the next scan starts from the beginning of the list.

## Fundamental Changes

- **`FDv2SourceResult` and `FDv2PayloadResponse`** now carry a fallback indicator alongside their existing data. All factory methods retain backward-compatible defaults.

- **`ModeDefinition`** supports a separate list of FDv1 fallback synchronizer factories, kept distinct from regular synchronizers so they can be independently blocked/unblocked.

- **The FDv2DataSource constructor** accepts an optional list of FDv1 fallback synchronizers. It handles wrapping them as blocked entries internally, so callers just pass plain factory lists.

- **`SourceManager.fdv1Fallback()`** is now a real implementation (previously a stub). It synchronizes on the active source lock to prevent races with concurrent synchronizer selection.

- **The default mode table** configures an FDv1 polling fallback for STREAMING, POLLING, and BACKGROUND modes. OFFLINE and ONE_SHOT don't have active synchronizers, so they have no fallback slot.

## Some Things That Aren't In Scope

- **Recovery from FDv1 back to FDv2** — Once the SDK falls back to FDv1, it stays there for the lifetime of the data source. Recovery would require server-side coordination that isn't specified yet.
- **FDv1 streaming fallback** — The spec allows for it, but only polling is implemented as the fallback synchronizer. This matches the js-core reference implementation.
- **Cache initializer** — The mode table has TODO placeholders for cache initializers; that's a separate piece of work.
- **State debouncing** — Network and lifecycle event debouncing (CONNMODE 3.5) is deferred to a future PR.

---

**Requirements**

**Related issues**

**Additional context**

---

> [!NOTE]
> **Medium Risk**
> Introduces new cross-cutting fallback behavior in the flag update pipeline (header propagation, mode tables, and synchronizer orchestration), which could affect initialization/state transitions and update delivery if mishandled.
> 
> **Overview**
> Adds support for **automatic FDv2→FDv1 fallback** when the backend returns `x-ld-fd-fallback: true`, switching the active synchronizer to an FDv1 polling implementation.
> 
> This threads an `fdv1Fallback` signal through `FDv2PayloadResponse`/`FDv2SourceResult`, teaches both polling (`DefaultFDv2Requestor`/`FDv2PollingBase`) and streaming (`FDv2StreamingSynchronizer`) paths to detect the header, and updates `FDv2DataSource`/`SourceManager` to honor it by blocking FDv2 synchronizers and activating a dedicated `FDv1PollingSynchronizer`.
> 
> It also extends `ModeDefinition`/`ResolvedModeDefinition` and the default mode table (`DataSystemComponents.makeDefaultModeTable`/`DataSystemBuilder`) to include an optional FDv1 fallback slot for STREAMING/POLLING/BACKGROUND, plumbs cache dir via `DataSourceBuildInputs`, moves poll interval defaults to `LDConfig`, and adds unit/contract-test coverage for header detection and fallback switching.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 16b9f8ff0b18ccc4106e2882653dd27a6f6973c3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/subsystems/FDv2SourceResult.java
```diff
@@ -74,21 +74,36 @@ public String getReason() {
...
     @NonNull
     public static FDv2SourceResult status(@NonNull Status status) {
-        return new FDv2SourceResult(SourceResultType.STATUS, null, status);
+        return new FDv2SourceResult(SourceResultType.STATUS, null, status, false);
+    }
+
+    @NonNull
+    public static FDv2SourceResult status(@NonNull Status status, boolean fdv1Fallback) {
+        return new FDv2SourceResult(SourceResultType.STATUS, null, status, fdv1Fallback);
     }
```
aaron-zeisler, line 97: @tanderson-ld Throughout this pull request, functions have been overloaded to introduce this `fdv1Fallback` boolean parameter. Question: Because the code is brand new and unused in production , does it make sense to have the overloads? Or should we just have one version of each of these overloaded methods. On the other hand, if this fallback feature is really temporary, does it actually help our future selves to keep the versions of the functions without the `fdv1Fallback` parameter?

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/HeaderConstants.java
```diff
@@ -0,0 +1,8 @@
+package com.launchdarkly.sdk.android;
+
+/**
+ * HTTP header constants used by FDv2 data sources.
+ */
+abstract class HeaderConstants {
+    static final String FDV1_FALLBACK_HEADER = "x-ld-fd-fallback";
+}
```
aaron-zeisler, line 8: @tanderson-ld I created this file to mirror the code in java-core. Is this appropriate? Can we actually just import the code from java-core here? FYI: There's another file with FDv2-related constants called `StandardEndpoints.java` and that file exists both here in the Android SDK and over in java-core.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ModeDefinition.java
```diff
@@ -27,13 +27,23 @@ final class ModeDefinition {
...
+
+    ModeDefinition(
+            @NonNull List<ComponentConfigurer<Initializer>> initializers,
+            @NonNull List<ComponentConfigurer<Synchronizer>> synchronizers,
+            @NonNull List<ComponentConfigurer<Synchronizer>> fdv1FallbackSynchronizers
     ) {
         this.initializers = Collections.unmodifiableList(initializers);
         this.synchronizers = Collections.unmodifiableList(synchronizers);
+        this.fdv1FallbackSynchronizers = Collections.unmodifiableList(fdv1FallbackSynchronizers);
     }
```
aaron-zeisler, line 52: Same question here re: "does it make sense to keep the overloaded functions?"

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
```diff
@@ -58,39 +59,46 @@ public interface DataSourceFactory<T> {
...
-     *                                next synchronizer
-     * @param recoveryTimeoutSeconds  seconds before attempting to recover to the primary
-     *                                synchronizer
+     * @param evaluationContext          the context to evaluate flags for
+     * @param initializers               factories for one-shot initializers, tried in order
+     * @param synchronizers              factories for recurring synchronizers, tried in order
+     * @param fdv1FallbackSynchronizers  factories for FDv1 fallback synchronizers, or null if none;
+     *                                   these are appended after the regular synchronizers in a
+     *                                   blocked state and only activated when the server sends the
+     *                                   {@code x-ld-fd-fallback} header
```
aaron-zeisler, line 91: @tanderson-ld This `fdv1FallbackSynchronizers` parameter has been added. Another approach I was considering was for the caller (FDv2DataSourceBuilder) to work the fallback synchronizers into the existing `synchronizers` parameter, instead of creating this new parameter.

That approach has 2 side effects:
1. the builder would be responsible for the logic below on lines 116-121, and
2. the `synchronizers` parameter would then become `List<SynchronizerFactoryWithState> synchronizerFactories`.

Looking for your opinion on this arrangement vs the other approach. Thanks!

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/subsystems/FDv2SourceResult.java
tanderson-ld: If you can get rid of the overloads, I would do that.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
```diff
@@ -384,6 +401,15 @@ private void runSynchronizers(
...
                             }
+
+                            if (running
+                                    && result.isFdv1Fallback()
+                                    && sourceManager.hasFDv1Fallback()
+                                    && !sourceManager.isCurrentSynchronizerFDv1Fallback()) {
+                                logger.info("Server signaled FDv1 fallback; switching to FDv1 polling synchronizer.");
+                                sourceManager.fdv1Fallback();
+                                running = false;
+                            }
```
aaron-zeisler, line 438: Notes from code review:
- Will running ever be false here (at the beginning of the if statement)?

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv1PollingSynchronizer.java
```diff
@@ -0,0 +1,261 @@
...
+            }
+
+            resultQueue.put(result);
+        } catch (RuntimeException e) {
+            LDUtil.logExceptionAtErrorLevel(logger, e, "Unexpected exception in FDv1 polling synchronizer task");
+            resultQueue.put(FDv2SourceResult.status(FDv2SourceResult.Status.interrupted(e)));
+        }
+    }
+
+    private FDv2SourceResult doPoll() {
```
aaron-zeisler, line 106: Code review notes:
- Can we safely refactor the production FDv1 code to reuse this entire doPoll() function?
- Similarly, can we reuse the existing code on lines 87-101?

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/subsystems/FDv2SourceResult.java
aaron-zeisler: I've addressed this in my latest push 👍

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/HeaderConstants.java
aaron-zeisler: We chatted about this in Slack and decided we would keep it as is

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ModeDefinition.java
aaron-zeisler: Same resolution as above

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
aaron-zeisler: In my latest push, I reduced this parameter to support just one fallback synchronizer. That mirrors both the java-core implementation and the js-core implementation, and that's also called out in the specs.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
aaron-zeisler: I've added a comment to this block. There are a couple situations where `running` is false: during SHUTDOWN, and as a result of TERMINAL_ERROR.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv1PollingSynchronizer.java
aaron-zeisler: I haven't addressed this yet

aaron-zeisler:
@tanderson-ld In the very last commit (titled "persist fdv1 fallback state ..."), I addressed your concern about staying in "FDv1 fallback mode" when the platform state changes. Interestingly, this type of protection isn't present in the js-core implementation. I'll bring this up w/ Ryan tomorrow to see if he's aware of that gap, or if he chose not to make those changes for some reason.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSourceBuilder.java
```diff
@@ -90,31 +96,50 @@ private Map<ConnectionMode, ModeDefinition> makeDefaultModeTable() {
...
-                Collections.<ComponentConfigurer<Synchronizer>>emptyList()
+                Collections.<ComponentConfigurer<Synchronizer>>emptyList(),
+                null
         ));
         table.put(ConnectionMode.BACKGROUND, new ModeDefinition(
                 // TODO: Arrays.asList(cacheInitializer) — add once implemented
                 Collections.<ComponentConfigurer<Initializer>>emptyList(),
-                Collections.singletonList(backgroundPollingSynchronizer)
+                Collections.singletonList(backgroundPollingSynchronizer),
+                fdv1FallbackPollingSynchronizer
```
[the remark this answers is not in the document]
aaron-zeisler: I'm inclined to leave this as-is for now. The spec doesn't call out any details for what the fallback polling frequency should be, or if it should be adjusted upon a platform state change. The js-core implementation takes this same approach. The FDv1 fallback is (hopefully) a rare edge case that shouldn't last long: polling more frequently in this state might be beneficial, after we introduce a mechanism to recover and go back to FDv2.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/DefaultFDv2Requestor.java
```diff
@@ -177,11 +177,17 @@ public void onResponse(@NonNull Call call, @NonNull Response response) {
...
+        String value = response.header(HeaderConstants.FDV1_FALLBACK_HEADER);
+        return value != null && value.equalsIgnoreCase("true");
+    }
+
     private void handleResponse(
             @NonNull Response response,
             @NonNull LDAwaitFuture<FDv2PayloadResponse> future) {
         try {
             int code = response.code();
+            boolean fdv1Fallback = isFdv1Fallback(response);
```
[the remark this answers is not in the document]
aaron-zeisler: This aligns with the CSFDV2 spec (section 4.2.2), and it's addressed in the js-core implementation, so I'll fix it 👍

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv1PollingSynchronizer.java
aaron-zeisler: In my latest push, I've refactored this new FDv1PollingSynchronizer by injecting a FeatureFetcher object. That's the same helper class that the existing PollingDataSource uses 👍

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/DefaultFDv2Requestor.java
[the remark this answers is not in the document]
aaron-zeisler: I've addressed this in my latest push

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
```diff
@@ -267,6 +282,18 @@ private void runInitializers(
...
+                        && sourceManager.hasFDv1Fallback()) {
+                    logger.info("Server signaled FDv1 fallback during initialization; " +
+                            "skipping remaining initializers.");
+                    sourceManager.fdv1Fallback();
+                    if (anyDataReceived) {
+                        sink.setStatus(DataSourceState.VALID, null);
+                        tryCompleteStart(true, null);
+                    }
+                    return;
+                }
```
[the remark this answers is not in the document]
aaron-zeisler: Addressed this in my latest push

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSourceBuilder.java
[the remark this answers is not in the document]
tanderson-ld: I think let's get more input/discussion on this.

If we go into fallback for some extended period of time, this will be a lot of extra polls, consider the ratio of apps in the background vs in the foreground across the population of running applications (I'm guessing more are in background than foreground by a significant factor since most don't swipe kill apps) and then consider that foreground is polling many times faster than background polling frequency.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv1PollingSynchronizer.java
```diff
@@ -0,0 +1,195 @@
...
+     *                           false if the fetcher is shared and managed externally
+     * @param executor           scheduler for recurring poll tasks
+     * @param initialDelayMillis delay before the first poll in milliseconds
+     * @param pollIntervalMillis delay between the end of one poll and the start of the next
+     * @param logger             logger
+     */
+    FDv1PollingSynchronizer(
+            @NonNull LDContext evaluationContext,
+            @NonNull FeatureFetcher fetcher,
+            boolean ownsFetcher,
```
tanderson-ld: We should remove the ownsFetcher if possible.  I gather it was added for testing after searching around a bit.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv1PollingSynchronizer.java
```diff
@@ -0,0 +1,195 @@
...
+                    pollIntervalMillis,
+                    TimeUnit.MILLISECONDS);
+        }
+    }
+
+    private void pollAndEnqueue() {
+        try {
+            FDv2SourceResult result = doPoll();
+
+            if (result.getResultType() == com.launchdarkly.sdk.fdv2.SourceResultType.STATUS) {
```
tanderson-ld: Try to reduce the length of the type reference (here and elsewhere)

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv1PollingSynchronizer.java
```diff
@@ -0,0 +1,195 @@
...
+            } else {
+                LDUtil.logExceptionAtErrorLevel(logger, cause, "FDv1 fallback polling failed");
+            }
+            return FDv2SourceResult.status(FDv2SourceResult.Status.interrupted(cause), false);
+        } catch (SerializationException e) {
+            LDUtil.logExceptionAtErrorLevel(logger, e, "FDv1 fallback polling failed to parse response");
+            LDFailure failure = new LDFailure(
+                    "FDv1 fallback: invalid JSON response", e, LDFailure.FailureType.INVALID_RESPONSE_BODY);
+            return FDv2SourceResult.status(FDv2SourceResult.Status.interrupted(failure), false);
+        }
```
tanderson-ld, line 168: Is it possible to refactor these try/catch/if blocks to be more sequential with midline returns and minimize scope of each try / catch?  There are a lot of code smells in this try / catch block.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv1PollingSynchronizer.java
```diff
@@ -0,0 +1,195 @@
...
+                    Selector.EMPTY,
+                    flags,
+                    null,
+                    true);
+
+            return FDv2SourceResult.changeSet(changeSet, false);
+
+        } catch (InterruptedException e) {
+            return FDv2SourceResult.status(FDv2SourceResult.Status.interrupted(e), false);
+        } catch (java.util.concurrent.ExecutionException e) {
```
tanderson-ld:
```suggestion
        } catch (ExecutionException e) {
```

In general to reduce verbosity, types shouldn't call out the whole package path if there is no need for it.  A valid need would be disambiguation if two types are closely named.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv1PollingSynchronizer.java
```diff
@@ -0,0 +1,195 @@
...
+            Throwable cause = e.getCause() != null ? e.getCause() : e;
+
+            if (cause instanceof LDInvalidResponseCodeFailure) {
```
tanderson-ld: It is unexpected to see application level concerns `cause instanceof LDInvalidResponseCodeFailure` inside an execution level concern `catch (java.util.concurrent.ExecutionException e)`.  This indicates some funky-ness.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv1PollingSynchronizer.java
```diff
@@ -0,0 +1,195 @@
...
+                    flags,
+                    null,
+                    true);
+
+            return FDv2SourceResult.changeSet(changeSet, false);
+
+        } catch (InterruptedException e) {
+            return FDv2SourceResult.status(FDv2SourceResult.Status.interrupted(e), false);
+        } catch (java.util.concurrent.ExecutionException e) {
+            Throwable cause = e.getCause() != null ? e.getCause() : e;
```
tanderson-ld, line 165: Usually this sort of get cause to analyze a deeper exception isn't done.  The reason is that another layer could be introduced in the future and you start to get an onion of exceptions.  To adhere to separation of concerns, this layer shouldn't assume an exception is at a certain layer of the onion.  Is this already being done elsewhere for analyzing the LDInvalidResponseCodeFailure?

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv1PollingSynchronizer.java
```diff
@@ -0,0 +1,195 @@
...
+            if (scheduledTask != null) {
+                scheduledTask.cancel(false);
+                scheduledTask = null;
+            }
+        }
+        shutdownFuture.set(FDv2SourceResult.status(FDv2SourceResult.Status.shutdown(), false));
+        closeFetcher();
+    }
+
+    private void closeFetcher() {
```
tanderson-ld: I think if ownsFetcher goes away, this goes into close?  Also the spot that is calling closeFetcher needs to be updated then.

tanderson-ld (commented):
Still reviewing, but posting current comments at my end of day.  Will resume tomorrow.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
```diff
@@ -384,6 +417,20 @@ private void runSynchronizers(
...
+                            // synchronizer is still running (not shut down or terminally errored),
+                            // a fallback slot exists, and we aren't already on FDv1.
+                            if (running
+                                    && result.isFdv1Fallback()
+                                    && sourceManager.hasFDv1Fallback()
+                                    && !sourceManager.isCurrentSynchronizerFDv1Fallback()) {
+                                logger.info("Server signaled FDv1 fallback; switching to FDv1 polling synchronizer.");
+                                sourceManager.fdv1Fallback();
+                                running = false;
+                            }
```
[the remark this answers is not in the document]
aaron-zeisler: I've addressed this in my latest push

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv1PollingSynchronizer.java
aaron-zeisler: Done 👍

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv1PollingSynchronizer.java
aaron-zeisler: Done throughout the file 👍

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv1PollingSynchronizer.java
aaron-zeisler: Done

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv1PollingSynchronizer.java
aaron-zeisler: @tanderson-ld I attempted to address this comment and your previous comment (with a lot of help from Cursor). Please take a look at it - if it's still a mess, we may want to undo the commit where I introduced the FeatureFetcher object (commit 3fd905c). The goal of that commit is to reuse existing code, but it may have introduced more complexity than it saved.

tanderson-ld: Much better.  Can `catch (ExecutionException e)` just be `catch (Exception e)`?

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
tanderson-ld, line 245: Java core does not handle fallback in the initializers.  Who is right?  This PR or java core?  I can't remember if not handling it in initializers was an intentional choice.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
```diff
@@ -243,6 +258,12 @@ private void runInitializers(
                             // A non-empty selector means the payload is fully current; the
                             // initializer is done and synchronizers can take over from here.
                             if (!changeSet.getSelector().isEmpty()) {
```
tanderson-ld, line 283: Should this fallback check be conditional on the presence of a selector?  I don't think the selector should be looked at before the fallback concern is handled.

Perhaps this is just due to the midline return?  Still feels a little off.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
tanderson-ld: I see the JS impl is handling it in runInitializers.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSourceBuilder.java
```diff
@@ -90,31 +90,45 @@ private Map<ConnectionMode, ModeDefinition> makeDefaultModeTable() {
                     0, LDConfig.DEFAULT_BACKGROUND_POLL_INTERVAL_MILLIS, ctx.getBaseLogger());
         };
 
+        ComponentConfigurer<Synchronizer> fdv1FallbackPollingSynchronizer = ctx -> {
```
tanderson-ld: May need to merge main.  My PR that merged to main switched from ComponentConfigurer<T> to DataSourceBuilder<T> for Initializers and Synchronizers.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
aaron-zeisler: +1, js-core is handling fallback in the initializers, and I used that as a model. The spec is very general and doesn't explicitly call it out falling back in initializers vs synchronizers:

Requirement 4.2.2 (Response Headers):
> Client-side SDKs MUST read the "x-ld-fd-fallback" response header and, if "true", transition to FDv1 data sources.

Requirement 2.2.1 (FDv1 Fallback Endpoint Paths):
> Client-side SDKs MUST support fallback to FDv1 endpoints when the "x-ld-fd-fallback" response header is "true".

Requirement 9.1.1 (FDv2 to FDv1 Fallback):
> The SDK MUST fall back from FDv2 to FDv1 data sources when the "x-ld-fd-fallback" response header is "true".

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv1PollingSynchronizer.java
aaron-zeisler: Done

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSourceBuilder.java
aaron-zeisler: Merged! The merged includes a refactor to the FeatureFetcher class ... in order to get the CacheDir into the FeatureFetcher's constructor, I made a change to that "Inputs" object as well. After that, the rest of the merge was straightforward.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv1PollingSynchronizer.java
```diff
@@ -0,0 +1,191 @@
...
+        this.fetcher = fetcher;
+        this.logger = logger;
+
+        synchronized (taskLock) {
+            scheduledTask = executor.scheduleWithFixedDelay(
+                    this::pollAndEnqueue,
+                    initialDelayMillis,
+                    pollIntervalMillis,
+                    TimeUnit.MILLISECONDS);
+        }
```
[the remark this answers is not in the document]
aaron-zeisler: I think this problem is theoretical only. In practice, when the fallback synchronizer is in use, all the other initializers and synchronizers will be closed. In other words, only one synchronizer runs at a time.

Regarding the statement "... especially during the transition period when both FDv2 and FDv1 components are active": The old synchronizer is fully closed before the new one is even built. There's no window where both are running simultaneously. I'm not sure what "transition period" is referring to.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
```diff
@@ -242,6 +257,12 @@ private void runInitializers(
...
+                                if (result.isFdv1Fallback()
+                                        && sourceManager.hasFDv1Fallback()) {
+                                    logger.info("Server signaled FDv1 fallback during initialization; " +
+                                            "switching to FDv1 synchronizer.");
+                                    sourceManager.fdv1Fallback();
+                                }
```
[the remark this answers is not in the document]
aaron-zeisler: Addressed in that last push

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
aaron-zeisler: I refactored this code based on what we talked about earlier today. I pulled the fallback handling up, which allowed me to remove some duplicated code, and no longer couples the selector stuff with the fallback stuff.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/DataSystemComponents.java
```diff
@@ -0,0 +1,250 @@
...
+    /**
+     * Returns a builder for a FDv1 polling synchronizer.
+     * <p>
+     * A FDv1 polling synchronizer periodically polls LaunchDarkly for feature flag updates using the FDv1 endpoints.
+     * The poll interval can be configured via
+     * {@link FDv1PollingSynchronizerBuilder#pollIntervalMillis(int)}.
+     *
+     * @return a FDv1 polling synchronizer builder
+     */
+    public static FDv1PollingSynchronizerBuilder fdv1PollingSynchronizer() {
```
tanderson-ld: We don't want this to be public since fdv1 fallback is an LD internal concern.  I don't think you necessarily have to put this in this class, but it should not be public if so.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv1PollingSynchronizer.java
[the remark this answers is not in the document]
tanderson-ld: Maybe see if you can figure out why Cursor bugbot thinks `and other synchronizers` is possible at the same time.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
```diff
@@ -383,6 +420,20 @@ private void runSynchronizers(
...
+                            // After processing the result, check whether the server signaled
+                            // that this environment should fall back to FDv1 (via the
+                            // x-ld-fd-fallback response header). We check regardless of
+                            // whether the synchronizer is still running — a terminal error
+                            // response can still carry the fallback header, and the server's
+                            // instruction to use FDv1 should take precedence.
+                            if (result.isFdv1Fallback()
+                                    && sourceManager.hasFDv1Fallback()
+                                    && !sourceManager.isCurrentSynchronizerFDv1Fallback()) {
+                                logger.info("Server signaled FDv1 fallback; switching to FDv1 polling synchronizer.");
```
tanderson-ld: Make private const for this message and commonize with other log location.

aaron-zeisler: Done

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/DataSystemComponents.java
aaron-zeisler: In that last push, I've moved makeDefaultModeTable() from DataSystemBuilder to DataSystemComponents. Then I updated the code so FDv1PollingSynchronizerBuilderImpl now implements DataSourceBuilder<Synchronizer> directly, instead of extending this FDv1PollingSynchronizerBuilder (which I deleted).

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv1PollingSynchronizer.java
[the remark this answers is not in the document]
aaron-zeisler: I asked Cursor to analyze this statement a few different ways. In each case, it seems like Cursor generally knew how the shared executor worked because of the comments in that class's definition. But when it made this comment on my PR, it hadn't taken one step further to understand exactly how many threads were being used, and how the threads are being juggled.

After writing this, it seems like adding a little bit more to the comments may help in the future.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/DataSystemComponents.java
```diff
@@ -102,6 +109,26 @@ public Synchronizer build(DataSourceBuildInputs inputs) {
...
+                inputs.getBaseLogger()
+            );
+            return new FDv1PollingSynchronizer(
+                inputs.getEvaluationContext(), fetcher,
+                inputs.getSharedExecutor(), 0,
+                PollingSynchronizerBuilder.DEFAULT_POLL_INTERVAL_MILLIS,
+                inputs.getBaseLogger()
+            );
+        }
+    }
```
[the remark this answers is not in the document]
aaron-zeisler: Same issue as above: [[link](https://github.com/launchdarkly/android-client-sdk/pull/339#discussion_r3019568924)]. @tanderson-ld our discussion here in github is open-ended, but I think we discussed this in our team meeting last Thursday, correct (or in Slack)?

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/integrations/DataSystemBuilder.java
```diff
@@ -243,67 +241,34 @@ public Map<ConnectionMode, ConnectionModeBuilder> getConnectionModeOverrides() {
...
 
         for (Map.Entry<ConnectionMode, ConnectionModeBuilder> entry : connectionModeOverrides.entrySet()) {
             ConnectionModeBuilder cmb = entry.getValue();
+
+            DataSourceBuilder<Synchronizer> fdv1FallbackSynchronizer = null;
+            if (!cmb.getInitializers().isEmpty() || !cmb.getSynchronizers().isEmpty()) {
+                fdv1FallbackSynchronizer = table.get(entry.getKey()).getFdv1FallbackSynchronizer(); // use fdv1 fallback from default mode table
+            } else {
+                fdv1FallbackSynchronizer = null;
+            }
```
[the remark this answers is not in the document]
aaron-zeisler: Done

tanderson-ld:
Looks good to me, just that one bugbot null check needed and benchtesting results.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSourceBuilder.java
```diff
@@ -176,7 +176,10 @@ private static ResolvedModeDefinition resolve(
         for (DataSourceBuilder<Synchronizer> builder : def.getSynchronizers()) {
             syncFactories.add(() -> builder.build(inputs));
         }
-        return new ResolvedModeDefinition(initFactories, syncFactories);
+        DataSourceBuilder<Synchronizer> fdv1FallbackSynchronizer = def.getFdv1FallbackSynchronizer();
+        FDv2DataSource.DataSourceFactory<Synchronizer> fdv1Factory =
+                fdv1FallbackSynchronizer != null ? () -> fdv1FallbackSynchronizer.build(inputs) : null;
+        return new ResolvedModeDefinition(initFactories, syncFactories, fdv1Factory);
```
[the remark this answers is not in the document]
aaron-zeisler: Just pushed a fix for this

tanderson-ld (approved):

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/integrations/DataSystemBuilder.java
```diff
@@ -243,67 +241,35 @@ public Map<ConnectionMode, ConnectionModeBuilder> getConnectionModeOverrides() {
...
         for (Map.Entry<ConnectionMode, ConnectionModeBuilder> entry : connectionModeOverrides.entrySet()) {
             ConnectionModeBuilder cmb = entry.getValue();
+
+            DataSourceBuilder<Synchronizer> fdv1FallbackSynchronizer = null;
+            if (!cmb.getInitializers().isEmpty() || !cmb.getSynchronizers().isEmpty()) {
+                ModeDefinition defaultForMode = table.get(entry.getKey());
+                fdv1FallbackSynchronizer = defaultForMode != null
+                        ? defaultForMode.getFdv1FallbackSynchronizer()
+                        : null;
+            }
```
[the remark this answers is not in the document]
aaron-zeisler: Note to self: See what js-core does here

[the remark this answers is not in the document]
aaron-zeisler: This was done by design. I've added a comment to explain.

Bugbot raises a fair point about an override that only contains initializers will get the fallback, even though no synchronizers are specified.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/DataSystemComponents.java
```diff
@@ -141,6 +177,63 @@ public static StreamingSynchronizerBuilder streamingSynchronizer() {
...
+                // TODO: cacheInitializer — add once implemented
+                Arrays.asList(/* cacheInitializer, */ pollingInitializer),
+                Arrays.asList(streamingSynchronizer, pollingSynchronizer),
+                fdv1FallbackPollingSynchronizerForeground
+        ));
+        table.put(ConnectionMode.POLLING, new ModeDefinition(
+                // TODO: Arrays.asList(cacheInitializer) — add once implemented
+                Collections.<DataSourceBuilder<Initializer>>emptyList(),
+                Collections.singletonList(pollingSynchronizer),
+                fdv1FallbackPollingSynchronizerForeground
```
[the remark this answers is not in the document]
aaron-zeisler: This is also by design. I think the alternative is to create two separate FDv1PollingSycnhronizer builders (and i believe only their polling durations would be different).

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2StreamingSynchronizer.java
```diff
@@ -278,7 +290,8 @@ void handleMessage(MessageEvent event) {
             resultQueue.put(FDv2SourceResult.status(
                     FDv2SourceResult.Status.interrupted(
                             new LDFailure("Failed to parse SSE event", e,
-                                    LDFailure.FailureType.INVALID_RESPONSE_BODY))));
```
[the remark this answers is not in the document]
aaron-zeisler: I don't know if the actual "ping" response will contain the fallback header. But when the SDK receives a "ping" message, it immediately polls to get data (per the spec), and the response from the poll is inspected for a fallback header. I think that's good enough.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/DataSystemComponents.java
```diff
@@ -102,6 +109,35 @@ public Synchronizer build(DataSourceBuildInputs inputs) {
...
+        @Override
+        public Synchronizer build(DataSourceBuildInputs inputs) {
+            FeatureFetcher fetcher = new HttpFeatureFlagFetcher(
+                inputs.getServiceEndpoints().getPollingBaseUri(),
+                inputs.isEvaluationReasons(),
+                inputs.getHttp().isUseReport(),
+                LDUtil.makeHttpProperties(inputs.getHttp()),
+                inputs.getCacheDir(),
+                inputs.getBaseLogger()
+            );
```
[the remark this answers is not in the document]
aaron-zeisler: Bugbot is confused between the base URI and the full path. The `pollingBaseUri` is really just the _base_ url, and it's the same between FDv1 and FDv2. Then the correct path is added to the baseUri when the `fetch()` function is called.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv1PollingSynchronizer.java
```diff
@@ -0,0 +1,191 @@
...
+    }
```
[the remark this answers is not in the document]
aaron-zeisler: This was our design choice. The FDv2 polling synchronizer behaves the same way.

aaron-zeisler:
@tanderson-ld The rest of the open Bugbot issues are either not actual issues, or they're comments about things that we did by design. This looks good to me 👍

PR #342, opened 2026-04-06
chore: FDv2 Cache Initializer

aaron-zeisler:
## Goal

Implement the FDv2 cache initializer so that cached flag data is loaded from local storage as the first step in every connection mode's initializer chain, providing immediate flag values while the network initializer fetches fresh data.

## Approach

The FDv1 production code loads cached flags synchronously during `ContextDataManager.switchToContext()`. In FDv2, this cache loading is modeled as a dedicated `Initializer` — the first in every mode's chain — per CONNMODE spec 4.1.1.

Cache initializers are identified by the `InitializerFromCache` marker interface applied to their builder. At mode-resolution time, `FDv2DataSourceBuilder` wraps those builders' factories so the marker is preserved through the factory indirection, and at construction time `FDv2DataSource` partitions its initializers into cache initializers and general initializers. Cache initializers run **synchronously on the caller's thread before executor dispatch**, matching FDv1 behavior where an SDK initializing with `timeout=0` still receives cached values immediately. General initializers (e.g. polling) then run on the executor as before.

The cache initializer's dependency on `ReadOnlyPerEnvironmentData` is wired through a new package-private `DataSourceBuildInputsInternal` class that extends the public `DataSourceBuildInputs`, following the same pattern as `ClientContext`/`ClientContextImpl`. Internal SDK components call `DataSourceBuildInputsInternal.get(inputs)` to unwrap internal-only properties, while the public API surface remains unchanged.

## Key Behaviors

- **Cache hit** returns a `CHANGE_SET` with `ChangeSetType.Full`, `Selector.EMPTY`, and `persist=false`, so the orchestrator applies the data immediately but continues to the next initializer for a verified selector from the server
- **Cache miss**, **missing persistent store**, and **exceptions during cache read** all return a `ChangeSetType.None` changeset — analogous to "transfer of none" / 304 Not Modified (CSFDV2 9.1.2). This signals "I checked the source and there is nothing new" rather than an error, so modes with no synchronizers (e.g. OFFLINE) still complete startup successfully on cache miss
- **Pre-startup execution:** cache initializers run synchronously before the executor-dispatched phase begins, guaranteeing cached flag values are applied before the SDK's startup timeout starts counting
- **Separation of concerns:** `runCacheInitializers()` applies changesets but ignores selectors/statuses and never completes start; `runGeneralInitializers()` is responsible for init completion and observes synchronizer availability via the existing `start()` orchestration
- **Redundant cache loads removed:** in the FDv2 path, `ContextDataManager` skips its prior cache-load-on-context-switch via a `skipCacheLoad` parameter on `switchToContext()`, keeping FDv1 behavior unchanged
- **`fdv1Fallback` is always `false`** since the cache is local (no server headers involved)

## Not In Scope

- Cache freshness tracking (CSFDV2 5.2.x)
- Data Availability = CACHED config option (CONNMODE 4.1.3) — whether cache alone completes initialization when synchronizers are available
- Poll interval relative to cache freshness (CSFDV2 5.3.9)

---

**Requirements**

**Related issues**

**Additional context**

---

> [!NOTE]
> **Medium Risk**
> Changes FDv2 startup/identify sequencing to apply cached flag data synchronously before executor-dispatched initialization, which can affect initialization timing and flag availability semantics across connection modes.
> 
> **Overview**
> Implements an FDv2 *cache initializer* that reads persisted flag data and applies it as the first initializer step in every default connection mode, returning `Full` changesets on cache hit and `None` on miss/no-store/errors.
> 
> Updates the FDv2 orchestration to **separate cache initializers from general initializers**, running cache initializers synchronously on the caller thread before `start()` dispatch, and adjusts success criteria to ignore `ChangeSetType.None` as “data received.”
> 
> Adds internal-only plumbing (`DataSourceBuildInputsInternal`, `ReadOnlyPerEnvironmentData`, and `InitializerFromCache`) to pass persistent-store access into built-in FDv2 components without expanding the public API, and avoids redundant cache loads by allowing `ContextDataManager.switchToContext(..., skipCacheLoad)` when FDv2 is in use (wired through `LDClient`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fc931a870daab7352c82a24865a2fa63b7bae21a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/subsystems/CachedFlagStore.java
```diff
@@ -0,0 +1,27 @@
...
+import java.util.Map;
+
+/**
+ * Provides read access to cached flag data for a specific evaluation context.
+ * <p>
+ * This interface bridges the persistence layer with FDv2 data source builders,
+ * allowing the cache initializer to load stored flags without depending on
+ * package-private types.
+ */
+public interface CachedFlagStore {
```
aaron-zeisler: Is it acceptable for this interface to be public? It's public because it's been added to the DataSourceBuildInputs class, which is public. As part of the "Inputs" class, it gets passed down into DataSourceBuilder.build(), where it's used to build the cache initializer.

aaron-zeisler: PerEnvironmentData is the concrete class that implements this interface.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2CacheInitializer.java
```diff
@@ -0,0 +1,99 @@
...
+ * <p>
+ * Per CONNMODE 4.1.2, the cache initializer returns data with {@code persist=false}
+ * and {@link Selector#EMPTY} (no selector), so the orchestrator continues to the next
+ * initializer (polling) to obtain a verified selector from the server. This provides
+ * immediate flag values from cache while the network initializer fetches fresh data.
+ * <p>
+ * A cache miss is reported as an {@link FDv2SourceResult.Status#interrupted} status,
+ * causing the orchestrator to move to the next initializer without delay.
+ */
+final class FDv2CacheInitializer implements Initializer {
```
aaron-zeisler, line 34: Here's a summary of how the spec requirements and the js-core implementation were used when developing this class:

| Decision | Source | Reasoning |
|---|---|---|
| `Selector.EMPTY` on cache result | CONNMODE 4.1.2 | Cache is unverified; empty selector tells the orchestrator to continue to the polling initializer for a real selector |
| `persist=false` on ChangeSet | CONNMODE 4.1.2 | Don't re-write data we just read from cache |
| Cache miss = `interrupted` | js-core pattern | Fast failure so the orchestrator immediately moves on; `interrupted` is the correct signal (not `terminalError`, which would stop the chain) |
| `fdv1Fallback=false` always | Logic | Cache is local storage, no server headers are involved |
| Nullable `cachedFlagStore` | Testing pragmatism | Test contexts don't have a persistent store; graceful degradation avoids test setup burden |

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/LDClient.java
```diff
@@ -497,11 +499,17 @@ private void identifyInternal(@NonNull LDContext context,
...
+        // times out or otherwise fails. This does not short-circuit initialization — the data
+        // source still performs its network request regardless.
+        if (config.dataSource instanceof FDv2DataSourceBuilder) {
+            // FDv2: just set the context; the FDv2CacheInitializer handles cache loading
+            // as the first step in the initializer chain.
+            contextDataManager.setCurrentContext(context);
+        } else {
+            // FDv1: load cached flags immediately while the data source fetches from the network.
+            contextDataManager.switchToContext(context);
+        }
```
aaron-zeisler: We might be able to replace the `if (config.dataSource instanceof FDv2DataSourceBuilder) {` with something from ConnectivityManager here. I think ConnectivityManager has an internal variable that describes "FDv1 or FDv2?", but that internal variable isn't exposed via a getter function. Let me know if you think that's cleaner.

aaron-zeisler:
@tanderson-ld  The 5th (and currently last) commit updates the code in LDClient and ContextDataManager. In FDv1, flags are loaded from the cache in ContextDataManager's constructor and also in LDClient's "identify" flow. But in FDv2, loading from cache in those places is redundant now that we have the cache initializer. I could defer this commit to a separate pull request if you feel that's cleaner.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/subsystems/CachedFlagStore.java
aaron-zeisler: From discussion:

Let's update this interface to align directly with PerEnvironmentData's getContextData() function. Let's move this interface into PersistentDataStoreWrapper.java (so it lives very close to PerEnvironmentData). Finally, does it make sense to just call this ReadOnlyPerEnvironmentData?

This means that the call sites will deal with `EnvironmentData` objects, so they'll need to be tweaked slightly.

aaron-zeisler: I've made this change in my latest push. The EnvironmentData class is private to the `android` package, but the "inputs" object is public in the `subsystems` package. So I decided to remove the cache from the "inputs" object and perform the dependency injection a different way.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/DataSystemComponents.java
```diff
@@ -202,32 +216,28 @@ public static Map<ConnectionMode, ModeDefinition> makeDefaultModeTable() {
...
-                // TODO: cacheInitializer and streamingInitializer — add once implemented
-                Arrays.asList(/* cacheInitializer, */ pollingInitializer /*, streamingInitializer, */),
+                // TODO: streamingInitializer — add once implemented
+                Arrays.asList(cacheInitializer, pollingInitializer /*, streamingInitializer */),
                 Collections.<DataSourceBuilder<Synchronizer>>emptyList(),
                 null
         ));
         table.put(ConnectionMode.BACKGROUND, new ModeDefinition(
-                // TODO: Arrays.asList(cacheInitializer) — add once implemented
-                Collections.<DataSourceBuilder<Initializer>>emptyList(),
```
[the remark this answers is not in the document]
aaron-zeisler: I've resolved this based on Todd and Ryan's input re: treating a cache miss like a "transfer of none".

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2CacheInitializer.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.concurrent.Future;
+
+/**
+ * FDv2 cache initializer: loads persisted flag data from the local cache as the first
+ * step in the initializer chain.
+ * <p>
+ * Per CONNMODE 4.1.2, the cache initializer returns data with {@code persist=false}
+ * and {@link Selector#EMPTY} (no selector), so the orchestrator continues to the next
+ * initializer (polling) to obtain a verified selector from the server. This provides
+ * immediate flag values from cache while the network initializer fetches fresh data.
```
tanderson-ld: Remove details about orchestration and other initializers, this class should not know they exist for separation of concerns.
```
 * Per CONNMODE 4.1.2, the cache initializer returns data with {@code persist=false}
 * and {@link Selector#EMPTY} (no selector).
```

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2CacheInitializer.java
```diff
@@ -0,0 +1,109 @@
...
+ * <p>
+ * A cache miss (or missing persistent store) is returned as a {@link ChangeSetType#None}
+ * changeset — analogous to "transfer of none" / 304 Not Modified (CSFDV2 9.1.2). This
+ * signals "I checked the source and there is nothing new" rather than an error, so the
+ * orchestrator records {@code anyDataReceived = true} and continues normally. This is
+ * critical for OFFLINE mode where no synchronizers follow: without it, a cache miss
+ * would leave the SDK in a failed initialization state.
```
tanderson-ld: This comment block has too much detail about other code that is not within this class's responsibility.  Revise to only talk about why it says transfer of none in that case.  Even 304 Not Modified is an HTTP status code and shouldn't be mentioned.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ContextDataManager.java
tanderson-ld: Consider just passing the skip cache load bool into this method.  Then you don't need to define the difference between "switch" and "set.  "setCurrentContext" wouldn't be necessary.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2CacheInitializer.java
```diff
@@ -0,0 +1,109 @@
...
+                        null,
+                        false);
+                logger.debug("Cache hit: loaded {} flags for context", flags.size());
+                resultFuture.set(FDv2SourceResult.changeSet(changeSet, false));
+            } catch (Exception e) {
+                logger.warn("Cache initializer failed: {}", e.toString());
+                resultFuture.set(FDv2SourceResult.status(
+                        FDv2SourceResult.Status.interrupted(e), false));
+            }
+        });
```
tanderson-ld: Something that might be worth testing.

Dispatching the execution onto the executor does have some overhead / delay associated with it as the other thread is scheduled to pick up the work.  LDFutures.anyOf also has dispatching internally.  If that overhead / delay is significant compared to the average time to read flags from the persistence, then this initializer may be better to just do the work right on the same thread as run() is called.  This would undermine the "close-ability" of this initializer, but if the interaction with persistence is fast enough, do you really need "close-ability?

We may need this code path to be as fast as possible because a client initing with a timeout of 0 expects to get cached values as quickly as possible.  We don't want to slow this use case down.

Unfortunately testing this on an emulator is not going to give good results so this would have to be done on a physical device.

This same concern does not apply to the polling or streaming initializers / synchronizers because the execution overhead is insignificant compared to the network latency (at least that is what we have been assuming).

tanderson-ld: Thinking more, we may need to do more extensive testing around this "hot path to cached flags" because the FDv2 data source is already using a fixed thread pool executor which also has some overhead to start the thread that will call run().

tanderson-ld:
We need to meet this requirement:

For SDKs that are inherently waiting for cache today even with a timeout of 0 seconds, we need to ensure that in FDv2 that same SDK major version continues to wait for the cache initializer.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2CacheInitializer.java
tanderson-ld: From my testing on my device, it looks like there is ~300us of delay.  From other benchmarking of the Android SDKs startup performance on my phone, the total time to get from init starting to fdv2Source_start_start is ~48ms, so this task delay is 0.5% of startup time.

So not a big impact from a speed standpoint.

Conclusion: try to run on thread that called run() if trivial to keep thread count down, but if there is any concern it is an issue, then don't change.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ContextDataManager.java
aaron-zeisler: Nice 👍 Thank you for the suggestion.

I've updated the code in that last push.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2CacheInitializer.java
aaron-zeisler: Addressed in that latest push

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2CacheInitializer.java
aaron-zeisler: Addressed in that latest push

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2CacheInitializer.java
```diff
@@ -0,0 +1,109 @@
...
+                        flags,
+                        null,
+                        false);
+                logger.debug("Cache hit: loaded {} flags for context", flags.size());
+                resultFuture.set(FDv2SourceResult.changeSet(changeSet, false));
+            } catch (Exception e) {
+                logger.warn("Cache initializer failed: {}", e.toString());
+                resultFuture.set(FDv2SourceResult.status(
+                        FDv2SourceResult.Status.interrupted(e), false));
+            }
```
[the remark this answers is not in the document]
aaron-zeisler: This should be fixed with my latest push.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2CacheInitializer.java
aaron-zeisler: This change was really straightforward. Updated in that latest push.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
```diff
@@ -362,9 +362,11 @@ private void runInitializers(
...
             initializer = sourceManager.getNextInitializerAndSetActive();
         }
-        // All initializers exhausted. If any gave us data (even without a final selector),
-        // consider initialization successful and let synchronizers keep the data current.
-        if (anyDataReceived) {
+        // All initializers exhausted. If data was received and no synchronizers will follow,
+        // consider initialization successful. When synchronizers are available, defer init
+        // completion to the synchronizer loop — the synchronizer is the authority on whether
+        // the SDK has a verified, up-to-date payload.
+        if (anyDataReceived && !sourceManager.hasAvailableSynchronizers()) {
```
tanderson-ld: I'm against this change without understanding the reason more.  `runInitializers` is designed to not know about synchronizers.  What is the reason?

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSourceBuilder.java
```diff
@@ -149,28 +152,40 @@ public void close() {
...
     ) {
         List<FDv2DataSource.DataSourceFactory<Initializer>> initFactories = new ArrayList<>();
         for (DataSourceBuilder<Initializer> builder : def.getInitializers()) {
-            initFactories.add(() -> builder.build(inputs));
+            // The cache initializer's dependency (ReadOnlyPerEnvironmentData) is only
+            // available at build time, not when the static mode table is constructed,
+            // so we inject it here by replacing the placeholder with a wired copy.
+            final DataSourceBuilder<Initializer> effective;
+            if (builder instanceof DataSystemComponents.CacheInitializerBuilderImpl) {
+                effective = new DataSystemComponents.CacheInitializerBuilderImpl(envData);
```
tanderson-ld: If `ReadOnlyPerEnvironmentData` was available in inputs, does this go away?

We had talked about a con with putting it in inputs which was: If we ever have customers implement initializers and synchronizers, then their code would get access to the readable env data which is undesirable.

We have a sort of "pattern" with ClientContext and ClientContextImpl that lets us pass around internal only information and "unwrap" it where necessary in internal components and we tell customers to not make such a call as it isn't protected by any contractual guarantees.  Could this sort of approach be used here?  We would have DataSourceBuildInputs and an extending class like DataSourceBuildInputsInternal or something like that.  ReadOnlyPerEnvironmentData would only be available on the Internal one.

We can discuss today as well.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
tanderson-ld, line 1: You had mentioned calling the cache initializer synchronously in start here right before `sharedExecutor.execute`.  Is that a pending change?

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSourceBuilder.java
aaron-zeisler: Yeah, if ReadOnlyPerEnvironemntData was available in the inputs, then this special handling for the cache initializer goes away. And that's correct: the con is that ReadOnlyPerEnvironmentData would need to be a public interface. And now that we've updated the interface's function to return EnvironmentData, I think EnvironmentData would need to be public as well? (The previous commit got around that last bit by returning a Map<> instead of an actual EnvironmentData object.)

What's the exact mechanism for the pattern you mentioned? Can you point out one instance of it being used? 🙏

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
aaron-zeisler: OK, let me look into this part again. Now that I've taken a step back, I agree it's weird to have this synchronizer-specific bit inside runInitializers().

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
aaron-zeisler: That change is in this other PR for now: https://github.com/launchdarkly/android-client-sdk/pull/346. I put it in the separate PR so it's easier to review on its own.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSourceBuilder.java
tanderson-ld: Any lines like `ClientContextImpl impl = ClientContextImpl.get(clientContext);`

So you could make a DataSourceBuildInputsInternal that would extend DataSourceBuildInputs. DataSourceBuildInputsInternal would have the ReadOnlyPerEnvironmentData member.

In the initializer that we write, we would receive the DataSourceBuildInputs and do something like

```
DataSourceBuildInputsInternal inputsInternal = DataSourceBuildInputsInternal.get(inputs)
inputsInternal.getPerEnvironmentDataIfAvailable()
```

This unwraps the inputs to get at the internal only properties.  A customer can do this in their own code, but we are free to break that contract as we make no guarantees.  I'm hand waving package visibility here, but I think DataSourceBuildInputs and DataSourceBuildInputsInternal having the same visibility isn't a problem.  Copious documentation on the DataSourceBuildInputsInternal class can provide warning to developers that it is for internal use only.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
aaron-zeisler: Updated in that last push

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSourceBuilder.java
aaron-zeisler: I was able to use this pattern 🎉  In the end, the changes were not that overwhelming, and it definitely simplifies the code in FDv2DataSourceBuilder 👍

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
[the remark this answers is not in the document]
aaron-zeisler: This is no longer an issue after the latest push.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2CacheInitializer.java
```diff
@@ -27,82 +26,81 @@
...
+                    ChangeSetType.None,
+                    Selector.EMPTY,
+                    Collections.emptyMap(),
+                    null,
+                    false), false);
+        }
 
-        return LDFutures.anyOf(shutdownFuture, resultFuture);
+        LDAwaitFuture<FDv2SourceResult> future = new LDAwaitFuture<>();
+        future.set(result);
```
aaron-zeisler, line 98: There's a built-in future type other than LDAwaitFuture that's already complete. Composable futures were added in Android 24.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
aaron-zeisler: That's all been merged into this pull request now, along with the changes that you made in your branch.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/subsystems/InitializerFromCache.java
```diff
@@ -0,0 +1,10 @@
+package com.launchdarkly.sdk.android.subsystems;
+
+import java.io.Closeable;
+import java.util.concurrent.Future;
```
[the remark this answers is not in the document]
aaron-zeisler: Resolved in that last push.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2CacheInitializer.java
tanderson-ld: This is still open.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
```diff
@@ -124,7 +139,8 @@ public interface DataSourceFactory<T> {
             allSynchronizers.add(fdv1);
         }
 
-        this.sourceManager = new SourceManager(allSynchronizers, new ArrayList<>(initializers));
+        // note that the source manager only uses the initializers after the cache initializers and not the cache initializers
+        this.sourceManager = new SourceManager(allSynchronizers, new ArrayList<>(generalInitializers));
```
tanderson-ld: new ArrayList isn't necessary (I wrote the code, sorry!)

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
```diff
@@ -165,12 +187,11 @@ public void start(@NonNull Callback<Boolean> resultCallback) {
                 }
 
                 if (sourceManager.hasInitializers()) {
-                    runInitializers(context, dataSourceUpdateSink);
+                    runGeneralInitializers(context, dataSourceUpdateSink);
                 }
 
                 if (!sourceManager.hasAvailableSynchronizers()) {
                     if (!startCompleted.get()) {
-                        // try to claim this is the cause of the shutdown, but it might have already been set by an intentional stop().
```
tanderson-ld, line 173: This comment doesn't need to be removed. It is still valid.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSourceBuilder.java
```diff
@@ -170,7 +174,11 @@ private static ResolvedModeDefinition resolve(
     ) {
         List<FDv2DataSource.DataSourceFactory<Initializer>> initFactories = new ArrayList<>();
         for (DataSourceBuilder<Initializer> builder : def.getInitializers()) {
-            initFactories.add(() -> builder.build(inputs));
+            if (builder instanceof InitializerFromCache) {
+                initFactories.add(new CacheInitializerFactory(() -> builder.build(inputs)));
+            } else {
+                initFactories.add(() -> builder.build(inputs));
+            }
```
tanderson-ld, line 188: Add a comment to help explain what this logic is doing and that it is dealing with the DataSourceBuilder (public API) to FDv2DataSource.DataSourceFactory<Initializer> "handholding"

tanderson-ld (approved):

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2CacheInitializer.java
aaron-zeisler: Question about this. Our build.gradle files define the min SDK value at 21. Here's an example: https://github.com/launchdarkly/android-client-sdk/blob/3f0066bafb28f45ff5b755360b3fadee50a7ec3e/launchdarkly-android-client-sdk/build.gradle#L18

Since the CompletableFuture class was introduced in SDK version 24, I think the SDK will throw errors when it's run on API versions 21-23. Is this a big problem? What do you think about this? THose versions are pretty old and I wonder if we can update our min SDK value.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
aaron-zeisler: Updated

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
aaron-zeisler: Updated

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSourceBuilder.java
aaron-zeisler: Updated

PR #344, opened 2026-04-10
chore: FDv2 State Debouncing

aaron-zeisler:
## Goal

Implement FDv2 state change debouncing per CONNMODE spec 3.5 so that rapid platform state changes (network, lifecycle) coalesce into a single mode reconciliation instead of triggering one data source teardown/rebuild per event.

## Approach

A new `StateDebounceManager` class accumulates network and foreground state changes and fires a single reconciliation callback after a configurable debounce window (default 1 second per CONNMODE 3.5.4). `ConnectivityManager` routes all platform state events (network availability, foreground/background) through this manager.

To minimize divergence between the FDv1 and FDv2 code paths, the same `StateDebounceManager` is used in both. FDv1 runs it in **immediate mode** (`debounceMs == 0`), where the callback fires synchronously on each state change with no timer — preserving FDv1's existing behavior while keeping a single, unified state-handling path.

The reconcile callback (`ConnectivityManager.handleStateChange`) reads current state directly from `PlatformState` rather than from the debouncer, so there is one source of truth for platform state and the same handler serves both FDv1 (immediate) and FDv2 (debounced) flows.

## Key Behaviors

* **Debounce window:** 1 second by default (CONNMODE 3.5.4); configurable via package-private `LDConfig.Builder.connectionModeStateDebounceMs()` for tests
* **Coalescing:** Each state change resets the timer; only the final accumulated state is reconciled (CONNMODE 3.5.2-3.5.3)
* **A→B→C→A dedup (CONNMODE 3.5):** When the timer fires, the debouncer compares the current raw `(network, foreground)` to the snapshot at last fire and suppresses the callback if the platform churned through transitions and ended up at the prior baseline within the window. FDv1 immediate mode bypasses this dedup intentionally, since (a) there is no debounce window for state to oscillate within and (b) immediate-mode callbacks can fire on different threads
* **`identify()` bypass (CONNMODE 3.5.6):** Identify does NOT participate in debounce. `StateDebounceManager.reset()` cancels any pending timer and re-seeds both the state mirrors and the A→B→C→A baseline atomically under one lock, so the new context starts with a clean window. The debounce manager is `final` and reused for the life of the `ConnectivityManager` rather than being destroyed and recreated, eliminating a swap-window race against listener callbacks
* **`identify()` after `shutDown()`:** `onContextChanged` short-circuits when the `ConnectivityManager` has already been shut down; the callback completes successfully without touching the closed debouncer or event processor and without constructing a data source that would be immediately abandoned
* **`setForceOffline()` bypass:** Legacy API remains immediate; safe because `resolveMode()` short-circuits to OFFLINE when force-offline is set, so any in-flight debounced callback resolves to the same mode. After `shutDown()`, `handleStateChange` short-circuits via a `closed` guard so a late `setForceOffline` cannot rebuild a data source on a torn-down `ConnectivityManager`
* **Event flushing:** When transitioning to background under FDv2, pending events are flushed synchronously _before_ entering the debounce window (CONNMODE 3.3.1) — the app may be about to be killed
* **Event processor updates are immediate:** `updateEventProcessor()` runs in the listener (not in the debounced callback) so events are correctly tagged as `inBackground` / `offline` even during the 1-second debounce window
* **Drain-on-close:** `StateDebounceManager.close()` cancels any pending timer and acts as a drain barrier — it does not return until any in-flight reconcile callback (FDv2 timer fire or FDv1 immediate-mode setter) has finished. This guarantees `ConnectivityManager.shutDown()` can tear down state without racing the reconcile callback. The manager uses two locks (`taskLock` for timer state, `workLock` for callback serialization), held in a fixed acquisition order so the reconcile callback can safely re-enter caller code that holds its own monitors

## Configuration

A new package-private `LDConfig.Builder.connectionModeStateDebounceMs(long)` setter allows instrumented tests to use a short debounce window (~50ms) so polling resumption tests still complete within their 2-second timeout. This is internal-only — end users get the spec-mandated 1-second default.

## Cleanup

* Removed unused `Debounce` class and the `pollDebouncer` field in `ConnectivityManager` — both were dead code from before SDK 4.0.0

## Not In Scope

* `setConnectionMode()` API support for runtime mode override (CONNMODE 3.5.5; tracked separately in SDK-1957)
* Cache freshness tracking and freshness-relative poll scheduling (CSFDV2 5.2.x, 5.3.9)
* User-configurable foreground/background mode selection (CONNMODE 2.2.2)
* Mode switch optimization when adjacent modes share the same data source config (CSFDV2 5.3.8)

---

**Requirements**

**Related issues**

**Additional context**

---

> [!NOTE]
> **Medium Risk**
> Changes core connectivity/lifecycle mode-switching behavior by introducing debounced reconciliation and new shutdown/identify semantics; errors could cause missed/extra data source rebuilds or timing regressions, though covered by new unit/instrumentation tests.
> 
> **Overview**
> Implements CONNMODE 3.5 state-change debouncing for FDv2 by routing foreground/network signals through a new `StateDebounceManager`, coalescing rapid transitions into a single reconciliation callback while keeping FDv1 behavior via a `debounceMs=0` immediate mode.
> 
> Updates `ConnectivityManager` to (a) flush events immediately on background transitions, (b) reset (not debounce) state tracking on `identify()`, (c) short-circuit context changes after shutdown, and (d) close/cancel pending debounce work during shutdown; it also adds a test-only config knob (`LDConfig.Builder.connectionModeStateDebounceMs`) to shorten the debounce window in tests.
> 
> Removes the old unused `Debounce` implementation and replaces/extends coverage with new `StateDebounceManagerTest`, expanded `ConnectivityManagerTest` debouncing/flush/shutdown cases, and updated Android instrumentation tests to use a shorter debounce.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 114c2db3ac471a5d122b12ac6d8dbc8291f46f25. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

[SDK-1957]: https://launchdarkly.atlassian.net/browse/SDK-1957?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/StateDebounceManager.java
tanderson-ld, line 1: Is it worth generalizing this to be able to handle N many properties of generic types?  Perhaps not trivial to identify which property is which?

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ConnectivityManager.java
```diff
@@ -570,13 +623,51 @@ private void updateEventProcessor(boolean forceOffline, boolean networkAvailable
...
+                debounceMs,
+                this::handleDebouncedModeStateChange
+        );
+    }
+
+    /**
+     * Reconciliation callback invoked by the {@link StateDebounceManager} when the debounce
+     * timer fires (CONNMODE 3.5.3). Reads the latest accumulated state from the debounce
+     * manager and triggers a data source update if the resolved mode has changed.
+     */
```
tanderson-ld: I see the FDv1 vs FDv2 concerns routing to different handle methods.  It seems like there would be a way to pass all calls through the debounce manager and just put him in "immediate mode", then the FDv1 vs FDv2 part just happens at construction.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/LDConfig.java
```diff
@@ -667,6 +677,12 @@ Builder persistentDataStore(PersistentDataStore persistentDataStore) {
             return this;
         }
 
+        // visible for testing — allows instrumented tests to use a shorter debounce window
+        Builder debounceMs(long debounceMs) {
```
tanderson-ld: Consider making this naming more specific, perhaps very specific.  connectionModeStateDebounceMs

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/StateDebounceManager.java
```diff
@@ -0,0 +1,94 @@
...
+        if (closed) {
+            return;
+        }
+        synchronized (lock) {
+            if (pendingTimer != null) {
+                pendingTimer.cancel(false);
+            }
+            pendingTimer = taskExecutor.scheduleTask(() -> {
+                if (!closed) {
+                    onReconcile.run();
```
tanderson-ld: I would expect the debounce to pass the debounced values along to the invocation so that all values must pass through the debouncer and it can ensure no out of order situations, but perhaps it doesn't matter?  It just seems a little odd that the runnable immediately calls getters on this class? I guess that they get latest then.  Not asking for any changes, just describing my observations.

Perhaps this is so the first state can be gotten without the delay?

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/StateDebounceManager.java
tanderson-ld: Upon looking at this again, the getters would be tricky.  Never mind!

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ConnectivityManager.java
```diff
@@ -170,21 +173,49 @@ public void shutDown() {
...
+                // CONNMODE 3.5.1: route through debounce window
+                StateDebounceManager dm = stateDebounceManager;
+                if (dm != null) {
+                    dm.setForeground(foreground);
+                }
+            } else {
+                // FDv1 path: handleModeStateChange updates event processor internally
+                handleModeStateChange();
             }
-            handleModeStateChange();
```
tanderson-ld, line 183: The two platform callbacks for network and foreground got very complicated.  I'm having trouble following them.  There is now a midline return, a debounce handle callback, and a regular return out the end of the function.

Consider sending all signals through the debounce manager, but it can be put in immediate mode, that should eliminate the top level `if (useFDv2ModeResolution)`  I think.

Does the eventProcess.flush() call need to be before the automatic mode switching midline return?

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ConnectivityManager.java
```diff
@@ -170,21 +173,49 @@ public void shutDown() {
                 ? ((FDv2DataSourceBuilder) dataSourceFactory).getResolutionTable()
                 : null;
 
+        if (useFDv2ModeResolution) {
+            this.stateDebounceManager = createDebounceManager();
```
tanderson-ld: I guess you don't need to make the debounce manager in the case where we are in fdv2, but automatic mode switching is off.

If you do make the debounce manager with "immediate mode" in FDv1 case, this region would need to consider that.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ConnectivityManager.java
```diff
@@ -570,13 +623,51 @@ private void updateEventProcessor(boolean forceOffline, boolean networkAvailable
...
+    /**
+     * Reconciliation callback invoked by the {@link StateDebounceManager} when the debounce
+     * timer fires (CONNMODE 3.5.3). Reads the latest accumulated state from the debounce
+     * manager and triggers a data source update if the resolved mode has changed.
+     */
+    private void handleDebouncedModeStateChange() {
+        StateDebounceManager dm = stateDebounceManager;
+        if (dm == null) {
+            return;
+        }
```
tanderson-ld: This null check indicates a design smell since we are in a function that implies debounce happened.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/StateDebounceManager.java
```diff
@@ -0,0 +1,94 @@
...
+                }
+            }, debounceMs);
```
tanderson-ld: Perhaps pendingTimer should be called lastTask?  I think it can be non-null even after the task runs.

Is it ok that the pendingTimer is not null even after the task runs?  Perhaps this is fine.

I'm NOT advocating for taking a lock inside the runnable and nulling it, that gets funky and creates a loop back to members of the debounce manager and that isn't good.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ConnectivityManager.java
```diff
@@ -204,6 +239,15 @@ void switchToContext(@NonNull LDContext context, @NonNull Callback<Void> onCompl
...
             onCompletion.onSuccess(null);
         } else {
+            // CONNMODE 3.5.6: identify bypasses debounce — close and recreate the manager
+            if (useFDv2ModeResolution) {
+                StateDebounceManager oldDm = stateDebounceManager;
+                if (oldDm != null) {
+                    oldDm.close();
+                }
+                stateDebounceManager = createDebounceManager();
+            }
```
tanderson-ld: Is the requirement that it resets the debounce window as well or just that identify does not get debounced?  It is possible to bypass it without resetting it.  Perhaps this is matching JS?

This makes me wonder if resetting somehow works around a race condition where identify comes in as the debounce manager fires its handleModeState.  I think this race condition would exist today in production code between identify and mode state change, so I don't think it is new.

aaron-zeisler: The spec (CONNMODE 3.5.6) says identify must not participate in debounce. Yeah, this matches the js-core implementation.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ConnectivityManager.java
```diff
@@ -66,7 +66,7 @@ class ConnectivityManager {
     private final TaskExecutor taskExecutor;
     private final boolean backgroundUpdatingDisabled;
     private final List<WeakReference<LDStatusListener>> statusListeners = new ArrayList<>();
-    private final Debounce pollDebouncer = new Debounce(); // FDv1 only
+
```
aaron-zeisler, line 68: I found that this `Debounce` class was actually not used throughout the codebase, so I've removed it.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ConnectivityManager.java
aaron-zeisler: In my latest push, I combined the feedback from a few of your comments here and created the "immediate mode" in the debounce manager. Then I was able to juggle both FDv1 and FDv2 with the debounce manager, and it simplified the state change handlers.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/LDConfig.java
aaron-zeisler: Yep, I've used your recommendation for naming in my latest push 👍 Thanks for being specific.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/StateDebounceManager.java
aaron-zeisler: I've updated the name in my latest push 👍

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/StateDebounceManager.java
aaron-zeisler: Addressed in the "immediate mode" commit

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ConnectivityManager.java
aaron-zeisler: Addressed in the "immediate mode" commit

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ConnectivityManager.java
aaron-zeisler: Addressed in the "immediate mode" commit

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ConnectivityManager.java
aaron-zeisler: Addressed in the "immediate mode" commit

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ConnectivityManager.java
```diff
@@ -170,21 +173,29 @@ public void shutDown() {
...
-            if (useFDv2ModeResolution && !autoModeSwitchingConfig.isLifecycle()) {
-                updateEventProcessor(forcedOffline.get(), platformState.isNetworkAvailable(), platformState.isForeground());
+            updateEventProcessor(forcedOffline.get(), platformState.isNetworkAvailable(), platformState.isForeground());
+            if (!autoModeSwitchingConfig.isLifecycle()) {
+                stateDebounceManager.trackForeground(foreground);
                 return;
             }
-            handleModeStateChange();
+            // CONNMODE 3.3.1: flush pending events before transitioning to background
+            if (useFDv2ModeResolution && !foreground) {
```
tanderson-ld: I think the flush when going to background is independent of if we are using FDv1 or FDv2.  If so, `useFDv2ModeResolution` can be removed from the conditional.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/StateDebounceManager.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    void close() {
+        closed = true;
+        synchronized (lock) {
+            if (lastTask != null) {
+                lastTask.cancel(false);
+                lastTask = null;
+            }
+        }
```
tanderson-ld, line 193: It is possible for onReconile to run after close() returns.

Example of the race condition
```
Interleaving:

Thread A — caller path: set* → resetTimer() reads closed → false.
Thread A — debounceMs == 0, so it falls into onReconcile.run() (no lock, no recheck).
Thread B — close() writes closed = true, acquires the lock, cleans up, returns.
Thread A — proceeds to execute onReconcile.run().
```

I think it is a requirement that the caller be assured onReconcile cannot run after close() returns.  The reason is that after close() is called, the ConnectivityManager is going to call updateDataSource and it shouldn't compete / be clobbered by onReconcile running.

Spot where close is called
```
            stateDebounceManager.close();
            stateDebounceManager = createDebounceManager();

            ModeState state = snapshotModeState();
            if (dataSource == null || dataSource.needsRefresh(!state.isForeground(), context)) {
                updateEventProcessor(forcedOffline.get(), state.isNetworkAvailable(), state.isForeground());
                updateDataSource(true, state, onCompletion);
            } else {
                onCompletion.onSuccess(null);
            }
```

tanderson-ld: Also a race in the task lambda down below

```
Executor has begun running the lambda. It evaluates if (!closed) → true.
Thread B — close() writes closed = true, enters the lock. The task is already running, so cancel(false) is a no-op. Returns.
Lambda — calls onReconcile.run() after close() has returned.
```

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ConnectivityManager.java
tanderson-ld, line 1: Some code paths getting state from platform state and some getting it from debounce manager risks a divergence and bug.  We shouldn't duplicate data if not necessary.  I think removing the state from debounce manager greatly reduces this risk.

tanderson-ld: switchToContext is replaced by a listener mechanism on `main`.  Need to merge `main` and resolve this.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/StateDebounceManager.java
```diff
@@ -0,0 +1,166 @@
...
+        resetTimer();
+    }
+
+    void setForeground(boolean fg) {
+        if (this.foreground == fg) {
+            return;
+        }
+        this.foreground = fg;
+        resetTimer();
+    }
```
aaron-zeisler, line 90: From code review: For immediate mode, don't have the logic in resetTimer(), because immediate mode doesn't need the timer at all. Put the immediate mode logic in these setter functions instead.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/StateDebounceManager.java
```diff
@@ -0,0 +1,166 @@
...
+            // FDv1 immediate mode: fire synchronously, bypass dedup. See class
+            // Javadoc for why dedup is intentionally skipped here.
+            onReconcile.run();
+            return;
+        }
+        synchronized (lock) {
+            if (lastTask != null) {
+                lastTask.cancel(false);
+            }
+            lastTask = taskExecutor.scheduleTask(this::fireIfChanged, debounceMs);
```
aaron-zeisler: From the code review: there's a race condition where fireIfChanged is running when close() is called. So a synchronizer could start while the StateDebounceManager is shutting down.
Does this race condition exist on main?

tanderson-ld (approved):

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/ConnectivityManager.java
aaron-zeisler: Thanks, I've addressed this in the latest commit. This also resolved Bugbot's issue.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/StateDebounceManager.java
tanderson-ld: Resolved by two lock solution from pairing.

PR #346, opened 2026-04-14
chore: FDv2 Cache Initializer: Guaranteed cache load before the startup timeout begins

aaron-zeisler:
## Summary

Guarantees that FDv2 cached flag data is available before the startup timeout begins, matching FDv1 behavior where cache was loaded synchronously in `ContextDataManager`'s constructor. Previously, the FDv2 cache initializer ran asynchronously on the executor thread alongside the timeout, meaning clients using `timeout=0` could miss cached data entirely. Now, `FDv2DataSource.start()` executes a two-pass initializer strategy: pre-startup initializers (identified by `isRequiredBeforeStartup()`) run synchronously on the calling thread before the executor is dispatched, while remaining initializers run on the executor as before. Both passes reuse the existing `runInitializers()` method with no changes to result-handling logic.

To support this, initializers are now built eagerly in `FDv2DataSourceBuilder.resolve()` instead of being wrapped in lazy factory lambdas, and `SourceManager` accepts pre-built `List<Initializer>` with a filtered `getNextInitializerAndSetActive(boolean isRequiredBeforeStartup)` method. `FDv2CacheInitializer` overrides the new `isRequiredBeforeStartup()` default method to return `true`. All existing tests are updated for the new signatures, and new tests verify eager execution threading, both passes running, OFFLINE cache miss behavior, and immediate data availability with zero timeout.

**Requirements**

**Related issues**

**Additional context**

---

> [!NOTE]
> **Medium Risk**
> Changes FDv2 startup sequencing and threading by running some initializers on the calling thread and eagerly constructing initializer instances, which could affect initialization timing and ordering. Behavior is covered by updated and new tests but touches core data source startup logic.
> 
> **Overview**
> Ensures cached flag data can be applied *before* the FDv2 startup timeout starts by introducing an initializer “eager pass” in `FDv2DataSource.start()` that runs `Initializer`s with `isRequiredBeforeStartup()` synchronously, then runs remaining initializers on the executor as before.
> 
> Refactors initializer wiring to support this: `Initializer` gains `isRequiredBeforeStartup()` (default `false`), `FDv2CacheInitializer` marks itself eager, and the builder/`ResolvedModeDefinition`/`SourceManager` switch from initializer factories to pre-built initializer instances with filtered iteration and an index reset between passes. Tests are updated accordingly and add coverage for eager vs deferred threading and zero-timeout/offline cache-miss scenarios.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e8caa9fc0851335940529eb7bcb79fca05ac798f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
+++ b/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
@@ -69,7 +69,7 @@ final class FDv2DataSource implements DataSource {
      */
     FDv2DataSource(
             @NonNull LDContext evaluationContext,
-            @NonNull List<DataSourceFactory<Initializer>> initializers,
+            @NonNull List<Initializer> initializers,
             @NonNull List<DataSourceFactory<Synchronizer>> synchronizers,
             @Nullable DataSourceFactory<Synchronizer> fdv1FallbackSynchronizer,
             @NonNull DataSourceUpdateSinkV2 dataSourceUpdateSink,
@@ -84,7 +84,7 @@ final class FDv2DataSource implements DataSource {
 
     /**
      * @param evaluationContext          the context to evaluate flags for
-     * @param initializers               factories for one-shot initializers, tried in order
+     * @param initializers               pre-built initializers, tried in order
      * @param synchronizers              factories for recurring synchronizers, tried in order
      * @param fdv1FallbackSynchronizer   factory for the FDv1 fallback synchronizer, or null if none;
      *                                   appended after the regular synchronizers in a blocked state
@@ -101,7 +101,7 @@ final class FDv2DataSource implements DataSource {
      */
     FDv2DataSource(
             @NonNull LDContext evaluationContext,
-            @NonNull List<DataSourceFactory<Initializer>> initializers,
+            @NonNull List<Initializer> initializers,
             @NonNull List<DataSourceFactory<Synchronizer>> synchronizers,
             @Nullable DataSourceFactory<Synchronizer> fdv1FallbackSynchronizer,
             @NonNull DataSourceUpdateSinkV2 dataSourceUpdateSink,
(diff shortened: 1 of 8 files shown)


launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
```diff
@@ -101,7 +101,7 @@ public interface DataSourceFactory<T> {
      */
     FDv2DataSource(
             @NonNull LDContext evaluationContext,
-            @NonNull List<DataSourceFactory<Initializer>> initializers,
+            @NonNull List<Initializer> initializers,
```
tanderson-ld, line 104: We still want factories here so we don't make initializers that are never reached.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/subsystems/Initializer.java
```diff
@@ -26,4 +26,18 @@ public interface Initializer extends Closeable {
...
+     * Eager initializers run synchronously on the calling thread during
+     * {@code FDv2DataSource.start()}, before work is dispatched to the executor.
+     * This guarantees their data is available even with a timeout of zero,
+     * matching the FDv1 behavior where cached data was loaded in the constructor.
+     *
+     * @return {@code true} if this initializer must complete before the timeout starts
+     */
+    default boolean isRequiredBeforeStartup() {
+        return false;
+    }
```
tanderson-ld, line 42: This is an internal concern and shouldn't be exposed on the interface.  If we make this public and have customers implement it, we can't claw it back.

An alternative would be a marker interface that specific initializers use and the FDv2 source knows how to check that marker interface.

Could also mark them with a boolean in the structure.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/FDv2DataSource.java
aaron-zeisler: I was hesitant about this. Devil's Advocate: won't all the initializers be executed? If one returns an error, the next one will be invoked until all are successful or are have failed.

Ah, the chain will stop when we have a selector. So it is possible to create a configuration with an initializer after a successful polling or streaming initializer runs.

I did this because the current indicator of "isRequiredBeforeStartup" is on the concrete, built instance. I'll go back to factories here once we identify a better way to mark initializers.

tanderson-ld: If a customer configures N initializers, but 99% of the time only the first two ever run, we spent time allocating and setting up the N-2 latter initializers.  A customer may do some interesting things in a constructor of a customer initializer (whenever we support that).

PR #374, opened 2026-07-09
fix: coalesce SharedPreferences writes to reduce memory pressure

tanderson-ld:
## Summary

- `setValue`/`setValues`/`clear` previously used `SharedPreferences.Editor.apply()`. Under bursts of context-data updates (a large segment cascade producing hundreds of stream PATCHes), each call enqueued a `MemoryCommitResult` on Android's `QueuedWork` — retaining a full forked map per snapshot. With a large flag payload this pushed the process past the Low Memory Killer threshold, terminating the app.
- Introduce in-memory coalescing at the `SharedPreferencesPersistentDataStore` layer. Writes fold into a single per-namespace `State` (clear / fullyDelete / changes). A background flush thread drains via `commit()` on a private single-threaded executor. Reads layer through `pending → committing → SharedPreferences` so `setValue`-then-`getValue` sees the latest write immediately.
- Peak retention under a burst is now `O(namespaces × unique keys touched)` regardless of how many `setValue` calls arrive.

## Design notes

- Uses `commit()` rather than `apply()`. The FLUSH_EXECUTOR is fsync-throttled, which naturally paces disk I/O without accumulating snapshots. Effective disk-write rate matches AOSP `SharedPreferencesImpl`'s post-hoc generation-skip coalescing under identical patterns.
- Failed commits are logged and abandoned rather than retried in a busy loop. Implicit recovery is provided by `SharedPreferences` itself: every `commit()` serializes the full in-memory map, so a failed write is picked up by the next successful commit for the same namespace — the same semantics `apply()` had.
- `clear` + `setValue` coalesce into one commit via `Editor.clear().putString(...).commit()` (AOSP processes `mClear` before `mModified`). Halves I/O for the clear+writes case and closes a partial-persist window where clear could land but changes couldn't.
- `fullyDelete`'s XML file removal fires regardless of the clear-commit outcome. Matches the pre-existing `apply()`-based behavior; strictly better on the file-recreation race because `commit()` is synchronous.
- A `@VisibleForTesting volatile Runnable preCommitRunnable` hook lets tests deterministically pause the drain thread between "state moved into committing" and "commit begins," enabling assertions on the layered read during in-flight commits.

## Tests

- 14 new tests covering `getKeys`, `clear(fullyDelete=true/false)` file-existence semantics, coalesced clear+setValue, layered `getValue` reads (pending vs disk, and same-key-across-layers precedence), concurrent 8×100 writes, and edge cases (empty-map `setValues`, `flushSynchronously` on an empty store).
- Two barrier-driven tests use the `preCommitRunnable` hook to hold the drain thread between the pending → committing move and the disk commit, then read the same key with distinct values in each of pending/committing/disk to catch layer-precedence bugs.
- One existing test (`valuePersistsAcrossInstances`) updated to call `flushSynchronously()` — writes now propagate to disk asynchronously.

## Test plan

- [ ] `./gradlew :launchdarkly-android-client-sdk:compileDebugJavaWithJavac :launchdarkly-android-client-sdk:compileDebugAndroidTestJavaWithJavac` passes.
- [ ] Full androidTest suite green on an emulator/device.
- [ ] Manual repro of the OOM scenario against `hello-android` with a large flag payload — confirm heap does not accumulate `MemoryCommitResult` snapshots.

---

> [!NOTE]
> **Medium Risk**
> Changes default persistence semantics (async disk vs immediate `apply`) and layered read/clear behavior across the SDK’s default store; well-tested but affects flag/context persistence under load.
> 
> **Overview**
> **`SharedPreferencesPersistentDataStore` no longer calls `Editor.apply()` on every write.** Writes are folded into per-namespace in-memory `State` (changes, clear, `fullyDelete`), drained on a shared single-thread executor with one `commit()` per namespace, and reads go through **pending → committing → disk** so recent writes are visible before flush.
> 
> **`getKeys` and `getAllNamespaces`** now overlay pending/committing state on top of disk (and union in-memory namespaces for `getAllNamespaces`). **`clear`** replaces pending state for that namespace and can coalesce with later `setValue` in one commit; **`fullyDelete`** file removal is deferred until a clear-only commit when appropriate.
> 
> Tests add **`flushSynchronously()`** and a **`preCommitRunnable`** hook for layer-precedence and concurrency coverage; **`valuePersistsAcrossInstances`** flushes before asserting cross-instance disk visibility.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 071dde1782e5aff6db68ab92116405940c0c5348. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/SharedPreferencesPersistentDataStore.java
+++ b/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/SharedPreferencesPersistentDataStore.java
@@ -252,21 +252,38 @@ final class SharedPreferencesPersistentDataStore implements PersistentDataStore 
     }
 
     /**
-     * Loops: move pending into committing, commit each namespace's state, remove from
-     * committing, check pending again. Exits when pending is empty. Each iteration
-     * starts with {@code committing} empty, since we remove entries after every commit.
+     * Loops: if committing has any leftover state (from a prior drain aborted by an
+     * Error inside commitState), drain it first; otherwise move pending into
+     * committing. Commit each namespace's state, remove from committing, repeat.
+     * Exits when both pending and committing are empty.
+     * <p>
+     * Draining committing before pulling from pending is intentional. A naïve
+     * {@code putAll(pending)} would overwrite any stranded entry whenever a writer
+     * arrived for the same namespace, silently losing the earlier coalesced writes.
+     * By processing committing first, its writes reach SharedPreferences'
+     * {@code mMap} before the newer pending writes are applied on top — natural
+     * merge, no State-level merge logic required.
      */
     private void drainPending() {
         try {
             while (true) {
                 Set<String> namespacesToCommit;
                 synchronized (lock) {
-                    if (pending.isEmpty()) {
-                        flushScheduled = false;
-                        return;
+                    if (committing.isEmpty()) {
+                        if (pending.isEmpty()) {
+                            flushScheduled = false;
+                            return;
+                        }
+                        // committing is empty here, so putAll cannot overwrite a
+                        // stranded entry from a prior failed drain.
+                        committing.putAll(pending);
+                        pending.clear();
                     }
-                    committing.putAll(pending);
-                    pending.clear();
+                    // If committing was non-empty on entry, drain it first without
+                    // pulling from pending. This preserves any state stranded by an
+                    // Error inside commitState — otherwise a naïve putAll(pending)
+                    // would overwrite it whenever a writer arrived for the same
+                    // namespace, silently losing the earlier coalesced writes.
                     namespacesToCommit = new HashSet<>(committing.keySet());
                 }
 
--- a/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/SharedPreferencesPersistentDataStore.java
+++ b/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/SharedPreferencesPersistentDataStore.java
@@ -82,6 +82,10 @@ final class SharedPreferencesPersistentDataStore implements PersistentDataStore 
     @VisibleForTesting
     volatile Runnable preCommitRunnable = null;
 
+    // Log-once-per-episode gate. Set when we catch a Throwable during persistence and
+    // decide to log it; cleared on the first successful commit that follows.
+    private boolean persistenceErrorLogged = false;
+
     public SharedPreferencesPersistentDataStore(Application application, LDLogger logger) {
         this.application = application;
         this.logger = logger;
@@ -308,6 +312,10 @@ final class SharedPreferencesPersistentDataStore implements PersistentDataStore 
             synchronized (lock) {
                 flushScheduled = false;
             }
+            if (!persistenceErrorLogged) {
+                logger.error("Encountered exception during persistence drain: {}", t);
+                persistenceErrorLogged = true;
+            }
             throw t;
         }
     }
@@ -321,11 +329,9 @@ final class SharedPreferencesPersistentDataStore implements PersistentDataStore 
      * writes — the net observable state (namespace contains only the new writes) is
      * unchanged.
      * <p>
-     * Any exception thrown from the underlying SharedPreferences plumbing (editor
-     * construction, putString, commit, or the file removal) is caught and logged.
-     * Letting an exception propagate would kill the enclosing drain runnable and leave
-     * {@code flushScheduled} stuck at {@code true}, blocking all future writes for the
-     * lifetime of this store instance.
+     * Any Throwable from the underlying SharedPreferences plumbing (editor construction,
+     * putString, commit, file removal) or from a test's preCommitRunnable is caught and
+     * (subject to log-once-per-episode gating) logged.
      */
     private void commitState(String namespace, State state) {
         try {
@@ -350,15 +356,21 @@ final class SharedPreferencesPersistentDataStore implements PersistentDataStore 
 
             editor.commit();
 
+            // Successful commit — clear the log-once gate
+            persistenceErrorLogged = false;
+
             if (state.clear && state.fullyDelete && state.changes.isEmpty()) {
                 File file = new File(application.getFilesDir().getParent()
                         + "/shared_prefs/" + namespace + ".xml");
                 //noinspection ResultOfMethodCallIgnored
                 file.delete();
             }
-        } catch (Exception e) {
-            logger.error("Encountered exception committing persistence for namespace '{}': {}",
-                    namespace, e);
+        } catch (Throwable t) {
+            if (!persistenceErrorLogged) {
+                logger.error("Encountered exception committing persistence for namespace '{}': {}",
+                        namespace, t);
+                persistenceErrorLogged = true;
+            }
         }
     }
 
--- a/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/SharedPreferencesPersistentDataStore.java
+++ b/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/SharedPreferencesPersistentDataStore.java
@@ -37,7 +37,7 @@ import java.util.concurrent.atomic.AtomicInteger;
  * state and writes each namespace's state to disk in a single
  * {@link SharedPreferences.Editor#commit()} call.
  * <p>
- * Reads resolve in a read through manner: pending → committing → {@link SharedPreferences}.
+ * Reads resolve in a read-through manner: pending, then committing, then {@link SharedPreferences}.
  * <p>
  * All public writing methods return promptly; the actual disk I/O happens on a shared
  * background executor via {@code commit()}. Multiple rapid operations on the same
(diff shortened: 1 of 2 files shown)


launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/SharedPreferencesPersistentDataStore.java
```diff
@@ -4,19 +4,84 @@
...
+ * to a single key.  SharedPreferences does not internally coalesce, so this improves
+ * its behavior.
+ * <p>
+ * Each namespace has a resolved {@link State}. Every incoming call is folded into the
+ * pending state for its namespace, producing a single resolved state regardless of how
+ * many calls were made. A background flush task drains the pending state into a committing
+ * state and writes each namespace's state to disk in a single
+ * {@link SharedPreferences.Editor#commit()} call.
+ * <p>
+ * Reads resolve in a read through manner: pending → committing → {@link SharedPreferences}.
```
kinyoklion: nit: It would be nice to de-unicode the arrows.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/SharedPreferencesPersistentDataStore.java
```diff
@@ -46,60 +126,201 @@ public String getValue(String storeNamespace, String key) {
...
+
+        // If a test has set a preCommitRunnable, run it now. Alternatives
+        // required more invasive mocking.
+        Runnable runnable = preCommitRunnable;
+        if (runnable != null) {
+            runnable.run();
+        }
+
+        try {
+            editor.commit();
```
kinyoklion, line 357: Do we care about the return value? I guess before we used apply and so that indicates that we are just doing a best effort.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/SharedPreferencesPersistentDataStore.java
```diff
@@ -46,60 +126,201 @@ public String getValue(String storeNamespace, String key) {
...
+                pending.clear();
+                namespacesToCommit = new HashSet<>(committing.keySet());
+            }
+
+            for (String namespace : namespacesToCommit) {
+                State state;
+                synchronized (lock) {
+                    state = committing.get(namespace);
+                }
+                commitState(namespace, state);
```
kinyoklion: Do we need to worry about runtime exceptions escaping commitState?

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/SharedPreferencesPersistentDataStore.java
tanderson-ld: Addressed in a434a616 — replaced with ASCII in both the class docstring and the drainPending docstring.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/SharedPreferencesPersistentDataStore.java
tanderson-ld: Addressed in a434a616 — the entire commitState body is now inside a single try/catch(Exception). Docstring calls out why: an escape would kill the drain runnable and leave flushScheduled stuck at true, deadlocking future writes for the instance's lifetime.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/SharedPreferencesPersistentDataStore.java
tanderson-ld: Right — this matches the prior apply()-based semantics. apply() returns void so there was never a return value to inspect; the file.delete() ran unconditionally when fullyDelete was set. Best-effort is the intended contract.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/SharedPreferencesPersistentDataStore.java
```diff
@@ -46,60 +126,201 @@ public String getValue(String storeNamespace, String key) {
...
+        try {
+            editor.commit();
+        } catch (Exception e) {
+            logger.error("Encountered exception committing persistence for namespace '{}': {}",
+                    namespace, e);
+        }
+
+        if (state.clear && state.fullyDelete && state.changes.isEmpty()) {
+            File file = new File(application.getFilesDir().getParent()
+                    + "/shared_prefs/" + namespace + ".xml");
```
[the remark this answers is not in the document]
tanderson-ld: Intentional — this matches the pre-fix behavior. The prior `apply()`-based clear() called `file.delete()` unconditionally when `fullyDelete` was set (apply() returns void, so there was never a way to inspect the commit result). Our new code preserves that semantic. On process restart after a failed clear-commit, SharedPreferences restores from the backup file — same durability window `apply()` had. Not a regression.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/SharedPreferencesPersistentDataStore.java
```diff
@@ -46,60 +126,201 @@ public String getValue(String storeNamespace, String key) {
...
+            }
+
+            for (String namespace : namespacesToCommit) {
+                State state;
+                synchronized (lock) {
+                    state = committing.get(namespace);
+                }
+                commitState(namespace, state);
+                synchronized (lock) {
+                    committing.remove(namespace);
```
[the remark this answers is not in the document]
tanderson-ld: Intentional — this is the "log-and-abandon, rely on implicit recovery" design documented in the class-level javadoc, matching the semantics of the prior `apply()`-based code.

Two things keep this safe:

1. `SharedPreferencesImpl.commitToMemory()` runs synchronously and always succeeds — only `writeToFile` can fail. So `mMap` already contains the coalesced values regardless of the disk-write outcome, and any subsequent `SharedPreferences.getString(...)` returns them. Runtime reads-your-writes is not broken.
2. `mDiskStateGeneration` is only advanced on a successful `writeToFile`. The next successful commit for the same namespace re-serializes the full `mMap` (SharedPreferences always writes the entire map, never deltas) and picks up any prior unwritten changes. So a failed write is implicitly recovered by any subsequent successful commit for that namespace.

The only durability window is process death between a failed write and the next successful commit for that namespace — which is the same window `apply()` had, since it doesn't fsync-block the caller either. An explicit retry loop would introduce a busy-loop hazard on persistent I/O failure (e.g., disk full) without adding real durability, since we can't hold writes indefinitely.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/SharedPreferencesPersistentDataStore.java
```diff
@@ -46,62 +126,209 @@ public String getValue(String storeNamespace, String key) {
...
+                }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
tanderson-ld: Good catch — the prior fix was scoped to commitState and didn't cover Errors (OOM, StackOverflow) or throwables originating in drainPending's own body (the HashSet allocation, HashMap ops). Addressed in bba5dbb8: wrapped drainPending in `try { ... } catch (Throwable t) { synchronized (lock) { flushScheduled = false; } throw t; }` so the flag resets on any abnormal exit and future writes can still schedule a drain. The rethrow preserves the original throwable for the executor's default handling.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/SharedPreferencesPersistentDataStore.java
```diff
@@ -46,62 +126,222 @@ public String getValue(String storeNamespace, String key) {
...
+     * starts with {@code committing} empty, since we remove entries after every commit.
+     */
+    private void drainPending() {
+        try {
+            while (true) {
+                Set<String> namespacesToCommit;
+                synchronized (lock) {
+                    if (pending.isEmpty()) {
+                        flushScheduled = false;
+                        return;
```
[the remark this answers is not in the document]
tanderson-ld: I think there is a limit to what we can do after such a failure, afterall, persistence is failing now.  Perhaps we should revise the behavior to stop trying to keep saving to shared preferences if there is an unexpected failure.  The downside to this is that our pending cache could keep growing.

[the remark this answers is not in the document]
tanderson-ld: Real find — the finding is valid for the Error path, and there was a second hazard I hadn't spotted: even when a later drain fires, `putAll(pending)` silently overwrites any stranded `committing` entry for the same namespace, losing the earlier coalesced writes. Addressed in 8ce78364: drain `committing` first when it's non-empty, only `putAll` from `pending` when `committing` is empty. That eliminates the overwrite entirely and lets SharedPreferences' `mMap` merge older and newer writes for us naturally (older State commits into mMap, newer State commits on top). Behavior unchanged on the happy path.

kinyoklion (approved):

PR #380, opened 2026-08-04
feat: add per-hook evaluation exposure deduplication

abelonogov-ld:
**Summary**

Opt-in deduplication of evaluation exposures for hooks. A hook is told about every evaluation until you wrap it, so nothing changes for existing hooks:

```java
Components.hooks()
    .addHook(new MetricsHook())                                    // every evaluation
    .addHook(new DedupingHook(new ObservabilityHook()))            // 10 minute window
    .addHook(new DedupingHook(new TelemetryHook(), 60_000))
    .addHook(new DedupingHook(new ExperimentHook(), myCustomDeduper))
```

New public API, all in `com.launchdarkly.sdk.android.integrations`:

- `HookDecorator` — an abstract hook that forwards every stage to the hook it wraps, and reports that hook's metadata. Decorators stack in either order.
- `DedupingHook` — a decorator that suppresses an evaluation whose result the wrapped hook has just been told about. Wrap with a window, or with a deduper of your own.
- `EvaluationExposureDeduper` — the policy. The window is its only setting. Subclass it to decide differently; `shouldRecord` and `reset` are the only methods `DedupingHook` calls.
- `EvaluationExposureKey` — what identifies an evaluation result: environment name, flag key, variation, flag version, experiment status, and fully qualified context key.
- `EvaluationSeriesContext.getEvaluationExposureKey()` — resolves that identity on demand, along with the `EvaluationExposureKeySupplier` the SDK hands the series context to resolve it with. An evaluation costs a flag lookup only when a hook asks, and only once however many hooks ask.

The policy keeps one record per flag per environment, holding the result that flag last reported. The wrapped hook hears about the flag again as soon as the result changes, and once per window while it stays the same. Tracking the last result rather than every result seen means a flag flipping back and forth cannot hide its flips, and it bounds the records to the flags the environment serves.

The decision is taken in `beforeEvaluation`, before the series opens, so a suppressed evaluation reaches neither stage of the wrapped hook. Hooks pair their stages: the observability plugin starts a span in the before stage and ends it in the after one, so suppressing only the after stage would leave that span in its map to be evicted later and exported with a meaningless duration and no `feature_flag` event.

`LDClient.identify` clears what the wrapped hook has been told about, so the first evaluation of each flag afterwards always reaches it. Analytics events are untouched: feature, debug, and summary events are still recorded for every evaluation, so the evaluation counts LaunchDarkly reports for a flag do not change.

- **A deduper configured on `Hook` itself.** It was the first shape this took. It puts a policy on the interface that every hook implementer has to think about, and it cannot be composed, so a hook that wanted deduplication plus anything else had nowhere to put the second behavior. The decorator gives both without touching `Hook`.
- **Filtering only `afterEvaluation`.** This is what the browser and React Native observability plugins do, and it is much less machinery: no exposure key before the evaluation, no supplier, no suppression marker in the series data. It is wrong here, because the mobile observability hook pairs its stages, as above.
- **Deduplicating on every distinct exposure seen, rather than one record per flag.** A flag that flips between two results would report neither flip after the first, and the set of remembered exposures grows with every result a flag has ever had.
- **Wall clock time for the window.** A correction that moves the device clock backwards leaves every recorded time in the future, so those flags stay suppressed until real time catches up, which for a large correction is hours of dropped exposures. Windows are measured against `SystemClock.elapsedRealtime()`, which no correction reaches and which, unlike `System.nanoTime()`, keeps counting while the device sleeps.

**Additional context**

- Additive and off by default. An unwrapped hook behaves exactly as it does today.
- A hook set on `LDConfig` is one instance shared by the clients for every environment in `secondaryMobileKeys`, and so is its deduper. The environment is therefore part of both the exposure key and the per-flag record; sharing a record across environments would make each look like the other having changed its result, and neither would ever be suppressed.
- Experiment status is a component of the key in its own right, because the version reported on events is the flag's own version: a prerequisite flipping can move an evaluation into or out of an experiment while it lands on the same variation of the same flag version.
- Mirrored on iOS in launchdarkly/ios-client-sdk, with docs in sdk-meta and ld-docs-private.

---

> [!NOTE]
> **Overview**
> Adds **opt-in deduplication** so wrapped hooks are not invoked again when repeated evaluations resolve to the **same exposure** within a configurable time window. Unwrapped hooks still see every evaluation.
> 
> Customers register **`DedupingHook`** around a hook (optional window or custom **`EvaluationExposureDeduper`**). **`EvaluationExposureKey`** (plus **`EvaluationSeriesContext.getEvaluationExposureKey()`**) identifies a result by environment hash, flag, returned value, variation, version, and context. Suppression happens in **`beforeEvaluation`**, so neither hook stage runs for repeats—important for hooks that pair before/after (e.g. observability spans). **`identify`** clears dedupe state; **analytics events are unchanged**.
> 
> **`LDClient`** now reads flag and context once per variation call and passes that read into **`HookRunner`** so exposure keys match what the evaluation returns. **`HooksConfigurationBuilder.addHook`** is added. The example app documents **`local.properties`** keys and shows live dedupe counters.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 435ab26a6febc9b16c1da4a0dfb83b850aeda9b2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


kinyoklion:
Some SDKs used to have functionality like this and it was removed. Not all data is available in the client as to which events matter when. Like an experiment iteration. So we don't dedupe events. We can offer ways and guidance about how to avoid these situations. And you are welcome to de-dupe obersavability data.

abelonogov-ld:
> Some SDKs used to have functionality like this and it was removed. Not all data is available in the client as to which events matter when. Like an experiment iteration. So we don't dedupe events. We can offer ways and guidance about how to avoid these situations. And you are welcome to de-dupe obersavability data.

@kinyoklion Sure, I am not going to dedup events at all, it should only dedup evaluation hook call. I will removes events from it. I feel this functionality where triggered by a customer story

tanderson-ld:
> it should only dedup evaluation hook call

I'm not sure I agree with this.  I think we should put this dedupe logic in as narrow a spot and as close to the consumer as possible since it does result in loss of information.  If you want one hook to get all evals and another to get deduped, configuring it at the top level doesn't work.

Can you make a hook decorator that does the deduping and you just wrap your hook in deduping decoration if you want it?

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/HookRunner.java
```diff
@@ -51,7 +66,7 @@ public void addHook(Hook hook) {
     }
 
     public EvaluationDetail<LDValue> withEvaluation(String method, String key, LDContext context, LDValue defaultValue, EvaluationMethod evalMethod) {
-        if (hooks.isEmpty()) {
+        if (hooks.isEmpty() || !exposureFilter.shouldReport(key, context)) {
```
tanderson-ld: This puts all hooks in the same bucket of filtering. If one hooks wants full fidelity signals and one wants filtered signals, this will tie them together and that won't be resolvable.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/integrations/Hook.java
```diff
@@ -33,6 +35,88 @@ public Hook(String name) {
...
+     *     Components.hooks()
+     *         .addHook(new ObservabilityHook().evaluationExposureDeduper())
+     * </code></pre>
+     *
+     * @return this hook
+     * @see #evaluationExposureDeduper(int, int)
+     */
+    public Hook evaluationExposureDeduper() {
+        return evaluationExposureDeduper(new EvaluationExposureDeduper());
+    }
```
tanderson-ld: I still prefer decoration with static helpers to increasing the amount of functionality in the Hook abstract class.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/EvaluationExposureKey.java
```diff
@@ -0,0 +1,35 @@
...
+     *
+     * @param environmentName the name of the environment being evaluated against
+     * @param flagKey the flag key
+     * @param variation the variation index of the result
+     * @param flagVersion the flag version reported on events
+     * @param inExperiment whether the evaluation was part of an experiment rollout
+     * @param fullyQualifiedContextKey the fully qualified key of the evaluation context
+     * @return a stable key identifying the evaluation result
+     */
+    static String of(String environmentName, String flagKey, int variation, int flagVersion,
```
tanderson-ld: Making a string for this isn't efficient is it?  Can't this just a be a class/struct depending on the language this will be done in?

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/integrations/EvaluationExposureDeduper.java
```diff
@@ -0,0 +1,170 @@
...
+     */
+    public synchronized void reset() {
+        lastRecordedAt.clear();
+    }
+
+    private void evict(long nowMillis) {
+        // Keys whose window has already elapsed no longer change the outcome of shouldRecord, so
+        // reclaim those first. They sort before any live key, so this stops at the first live one.
+        long cutoff = nowMillis - windowMillis;
+        for (Iterator<Map.Entry<String, Long>> it = lastRecordedAt.entrySet().iterator(); it.hasNext(); ) {
```
tanderson-ld: Feels like there should be a way to have a sorted index by recorded at and then you can do a binary search to find the ones that need to be evicted and then just delete everything on one side of that pivot point.  If this list has thousands of elements, and who knows what a customer might pass in for that size limit, I don't think a linear approach will cut it.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/integrations/Hook.java
```diff
@@ -33,6 +35,88 @@ public Hook(String name) {
...
+     *
+     * <pre><code>
+     *     Components.hooks()
+     *         .addHook(new ObservabilityHook().evaluationExposureDeduper(60_000, 2_000))
+     * </code></pre>
+     *
+     * @param windowMillis the dedupe window in milliseconds; zero or negative reports every
+     *                     evaluation
+     * @param maxSize the maximum number of exposure keys to track; zero or negative uses
+     *                {@link EvaluationExposureDeduper#DEFAULT_MAX_SIZE}
```
tanderson-ld: I'm not really a fan of the maxSize param.  The behavior of the SDK changes drastically depending on if you operate below maxSize, near maxSize, or well above maxSize.  There is also unexpected interactions between this number and the size of an environment and your use case, it seems impossible for a customer to pick an intelligent number.

Shouldn't this be some linear factor of the environment size in the worst case?  Each time one of the properties into the exposure key changes, all the entries that had that the old property can be dropped from the exposure tracking as we know such an exposure can never happen again.  It seems like the exposure tracking map is the same size as the environments flag count?

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/integrations/EvaluationExposureKey.java
```diff
@@ -0,0 +1,138 @@
...
+     * @param environmentName the name of the environment the evaluation was made against
+     * @param flagKey the flag key
+     * @param variation the variation index of the result
+     * @param flagVersion the flag version reported on events
+     * @param inExperiment whether the evaluation was part of an experiment rollout
+     * @param fullyQualifiedContextKey the fully qualified key of the evaluation context
+     */
+    public EvaluationExposureKey(String environmentName, String flagKey, int variation,
+                                 int flagVersion, boolean inExperiment,
+                                 String fullyQualifiedContextKey) {
```
tanderson-ld: What happens if one of these properties is not known and then becomes known?  For example, I believe environmentName is unknown for a period of time until init completes.  Evaluations can be occurring before init completes.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/integrations/EvaluationExposureDeduper.java
```diff
@@ -0,0 +1,177 @@
...
+
+    /**
+     * The result a flag last reported, and when. Mutable so that a flag costs one record for as long
+     * as it is tracked, however often its result changes.
+     */
+    private static final class LastReported {
+        private int variation;
+        private int flagVersion;
+        private boolean inExperiment;
+        private String fullyQualifiedContextKey;
```
tanderson-ld: Why doesn't LastReported just contain an EvaluationExposureKey and atMillis?  By reiterating all these properties (without flag key) we risk a silly mistake where one property gets added in the future in one spot, but not in another spot.

I am concerned with how many spots the group of properties that are relevant to this dedupe, are....duplicated.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/integrations/EvaluationExposureKey.java
```diff
@@ -0,0 +1,138 @@
...
+ * <p>
+ * Instances are immutable, and their hash code is computed once, the first time one is asked for.
+ */
+public final class EvaluationExposureKey {
+    private final String environmentName;
+    private final String flagKey;
+    private final int variation;
+    private final int flagVersion;
+    private final boolean inExperiment;
+    private final String fullyQualifiedContextKey;
```
tanderson-ld, line 49: I think these set of properties needs to be rigidly specified somewhere.  If SDK A has one set of properties and SDK B has another set of properties, the evaluation reporting will be different between them.  The specification would be the reference to keep SDKs consistent with each other.

I'm not sure if it warrants a complete spec in sdk-specs or just a confluence page.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/integrations/EvaluationExposureDeduper.java
```diff
@@ -0,0 +1,177 @@
...
+            return true;
+        }
+
+        TrackedFlag flag = new TrackedFlag(key);
+        LastReported reported = lastReported.get(flag);
+        if (reported == null) {
+            lastReported.put(flag, new LastReported(key, nowMillis));
+            return true;
+        }
+        if (reported.atMillis > nowMillis - windowMillis && reported.isSameResultAs(key)) {
```
tanderson-ld, line 92: The following seems more readable and has the midline return true which is more consistent with what you have above with other midline return trues.

```
// elapsed time greater than window or evaluation exposure key changed
if (nowMillis - reported.atMillis > windowMillis || !reported.isSameResultAs(key)) {
        reported.update(key, nowMillis);
        return true;
}

return false;
```

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/integrations/EvaluationExposureDeduper.java
```diff
@@ -0,0 +1,177 @@
...
+    /**
+     * The dedupe window used by a deduper built without a window of its own: 10 minutes, in
+     * milliseconds.
+     */
+    public static final int DEFAULT_WINDOW_MILLIS = 600_000;
+
+    private final long windowMillis;
+
+    // Holds one record per flag the application evaluates, in each environment it evaluates it in.
+    // Nothing is evicted, because that set is the flags the environment serves. Guarded by the
```
tanderson-ld: "nothing is evicted" is no longer accurate since reset() does evict them all.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/integrations/HookDecorator.java
tanderson-ld, line 1: What value does this class add?  I would get rid of it as it seems like premature commonization.  I think `getDelegate` is a decorator anti pattern as well since it invites developers to introspect on the decorator for logical checks, which is not usually a good thing.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/integrations/EvaluationExposureKey.java
abelonogov-ld: environmentName is never unknown and it is default on the start and it will be just in another bucket

abelonogov-ld: For other I put defaults if something missing

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/integrations/EvaluationExposureDeduper.java
abelonogov-ld: I believe the current way is good. It does some work in the end and early exits without doing work inside if is good pattern

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/LDClient.java
```diff
@@ -732,10 +732,16 @@ private EvaluationExposureKey exposureKey(EvaluationSeriesContext seriesContext,
...
-                flagVersion, inExperiment, seriesContext.context.getFullyQualifiedKey());
+        // The value the evaluation returns, which is the default value when there is no value to
+        // return: a flag the SDK has no data for, and a flag whose data carries no value, both fall
+        // back to it, as they do on the event the evaluation records. A value the calling method
+        // rejects as the wrong type also falls back to the default, but is not recognized here,
+        // because the type that method wanted is not part of the series context.
+        LDValue flagValue = flag == null ? LDValue.ofNull() : flag.getValue();
+        LDValue value = flagValue.isNull() ? seriesContext.defaultValue : flagValue;
+
+        return new EvaluationExposureKey(clientContextImpl.getEnvironmentName(), flagKey, value,
```
tanderson-ld: If we can use a hash of environment sdk key instead of environment name, that gets rid of the cloud dependency / environment name changing after we interact with the cloud.  Ultimately getting the name and updating the name really shouldn't be considered a new exposure anyways because the environment name has nothing to do with flag evals.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/integrations/EvaluationExposureKey.java
```diff
@@ -0,0 +1,182 @@
...
+
+    /**
+     * Identifies the environment the evaluation was made against, for comparison only: it is a hash
+     * of the mobile key, so nothing can be read out of it, and the SDK gives no guarantee about how
+     * it is derived beyond being the same for two evaluations made against the same environment and
+     * different otherwise.
+     *
+     * @return an opaque identifier for the environment the evaluation was made against
+     */
+    public String getEnvironmentId() {
```
tanderson-ld: Need to move away from `environmentId` to avoid confusion with https://github.com/launchdarkly/sdk-specs/blob/main/specs/HOOK-hooks/README.md#evaluationseriescontext 's field environmentId.  If these both have the same name, it will definitely lead to confusion.

tanderson-ld (approved):

PR #381, opened 2026-08-11
fix: Use a consistent evaluation context for hook invocations and prerequisite evaluations.

abelonogov-ld:
There are two primary purposes of this PR. 

1. Access the context once for a variation call and provide that context consistently to prerequisite calls and for the hook meta-data.
2. Reduce duplication through having an evaluation method that is re-used in each hook invocation site.

Currently the context and the store data are not transactional with each other. So there are limits to the approach, but it is generally an improvement regardless.

### DIFF
--- a/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/LDClient.java
+++ b/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/LDClient.java
@@ -620,21 +620,24 @@ public class LDClient implements LDClientInterface, Closeable {
     }
 
     /**
-     * Runs an evaluation, and the hooks around it, against one read of the flag.
+     * Runs an evaluation, and the hooks around it, against one read of the flag and of the evaluation
+     * context.
      * <p>
-     * The read is done here rather than left to the evaluation so that there is one place expressing
-     * the order of the read, the hooks and the evaluation, and so that what an evaluation is about to
-     * return can be described to a hook without reading the store a second time.
+     * Both are read here rather than left to the evaluation so that there is one place expressing the
+     * order of the reads, the hooks and the evaluation. Reading only the flag here and letting the
+     * evaluation re-read the context would leave an identify landing in between attributing the prior
+     * context's flag to the new context in events.
      */
     private EvaluationDetail<LDValue> evaluateWithHooks(String method, String key, LDValue defaultValue,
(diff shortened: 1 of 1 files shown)


tanderson-ld:
This seems like an improvement, but there are related cases that are not fixed (nor are they trivial to fix).  The related cases are a prerequisite graph where a diamond exists.  A -> B, A -> C, B -> D, C -> D.  I think even with this change, D will be read twice from the store.

A proper fix would be some sort of snapshot of storage, but that is much more involved.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/LDClient.java
```diff
@@ -553,137 +554,98 @@ public Map<String, LDValue> allFlags() {
...
     }
 
-    private EvaluationDetail<LDValue> variationDetailInternal(@NonNull String key, @NonNull LDValue defaultValue, boolean checkType, boolean needsReason) {
-        return variationDetailInternal(key, defaultValue, checkType, needsReason, null);
-    }
-
-    private EvaluationDetail<LDValue> variationDetailInternal(@NonNull String key, @NonNull LDValue defaultValue, boolean checkType, boolean needsReason, Set<String> visited) {
-        LDContext context = clientContextImpl.getEvaluationContext();
-        Flag flag = contextDataManager.getNonDeletedFlag(key); // returns null for nonexistent *or* deleted flag
+    private EvaluationDetail<LDValue> variationDetailInternal(@NonNull String key, @NonNull LDValue defaultValue, boolean checkType, boolean needsReason, Set<String> visited, @Nullable Flag flag, @NonNull LDContext context) {
```
tanderson-ld, line 648: Do you even need the flag key param now (the first param)?  It seems like this function just operates on the domain object Flag and not on the key since the lookup was done beforehand.

kinyoklion:
> This seems like an improvement, but there are related cases that are not fixed (nor are they trivial to fix). The related cases are a prerequisite graph where a diamond exists. A -> B, A -> C, B -> D, C -> D. I think even with this change, D will be read twice from the store.
> 
> A proper fix would be some sort of snapshot of storage, but that is much more involved.

Updated the PR description to note the limitations, but still highlight the value of the change.

kinyoklion:
This was superseded by PR 380 that contained all the same changes. So I am closing this PR.

PR #393, opened 2026-08-21
feat: add registerPlugin so a plugin can be added after the client starts

abelonogov-ld:
## Summary

Plugins could only be supplied through `LDConfig`, so an integration that learns about a plugin later — or that wants to instrument a client it did not configure — had no way in. This adds `LDClient.registerPlugin(Plugin)`, mirroring the method the .NET and Flutter SDKs already expose.

- **Hooks go live only once `register` succeeds.** This follows the .NET ordering, and it differs from configuration-time registration, where a plugin's hooks are active before `register` is called. The consequence is that a plugin whose registration failed never contributes hooks, and a plugin's own hooks do not observe the evaluations or identify calls its `register` makes.
- **`EnvironmentMetadata` is now retained on the instance** rather than living in a local map during `init`, so a plugin registered later is handed the same environment description as one configured up front. The `init` loop switches from a `Map<LDClient, EnvironmentMetadata>` to a `List<LDClient>` accordingly.
- **Failures are contained.** A throwing `getHooks` leaves the plugin unregistered; a throwing `register` yields a failure result passed to `onPluginsReady`. Neither propagates to the caller. Reading the plugin's name for the log message tolerates metadata that itself throws, so reporting one failure cannot raise another out of the handler reporting it.

Registration applies to the one client it is called on, so a multi-environment setup means calling it per environment.

## Test plan

Eight cases added to `LDClientPluginsTest`, all passing on an API 36 emulator alongside the three pre-existing plugin tests (11 total):

- [x] `registerPluginPassesClientAndEnvironmentMetadata`
- [x] `registerPluginActivatesBundledHooks`
- [x] `registerPluginDoesNotRunTheRegisteringPluginsOwnHooks`
- [x] `registerPluginReportsSuccessToOnPluginsReady`
- [x] `registerPluginDoesNotRegisterPluginWhoseGetHooksThrows`
- [x] `registerPluginToleratesRegisterThrowing`
- [x] `registerPluginRejectsNullPlugin`
- [x] `registerPluginAppliesOnlyToTheClientItIsCalledOn`
- [x] `testDebugUnitTest` green

---

> [!NOTE]
> **Overview**
> Adds **`LDClient.registerPlugin(Plugin)`** on `LDClientInterface` so integrations can attach plugins after init, aligned with other LaunchDarkly mobile SDKs. **`EnvironmentMetadata` is stored on each client** (replacing the init-only map) so late registration gets the same environment description as config-time plugins.
> 
> Plugin setup is **centralized in `registerPlugins`**, used from both init and `registerPlugin`: collect hooks via `getHooks`, **activate all hooks in one `HookRunner.addHooks` step**, then call `register`. Hooks are live **before** `register` runs (including work done inside `register`); a throwing `register` is logged and does not remove already-active hooks. `getHooks` failures skip registration; other plugins still register. **`registerPlugin` does not invoke `onPluginsReady`** (batch-only); init still notifies via `notifyPluginsReady`. **`Plugin.onPluginsReady` is deprecated** in Javadoc.
> 
> **`HookRunner`** gains batch `addHooks` (with `addHook` delegating to it) so a plugin’s hook set cannot be half-visible to concurrent evaluations. Tests cover runtime registration, multi-environment scoping, failure modes, and config-time behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c4e234d5f02ff433956e1c778a64a81fa7fda275. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/LDClient.java
+++ b/launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/LDClient.java
@@ -225,31 +225,12 @@ public class LDClient implements LDClientInterface, Closeable {
             instances = newInstances;
         }
 
-        // after instances have been created, set up hooks for each plugin of the instance and call register
+        // after instances have been created, register each instance's plugins
         for (LDClient instance : createdInstances) {
             EnvironmentMetadata metadata = instance.environmentMetadata;
 
-            for (Plugin plugin : instance.plugins) {
-                // try is for each plugin so that if one plugin has an issue, the others will have an opportunity to be used
-                try {
-                    List<Hook> pluginHooks = plugin.getHooks(metadata);
-                    for (Hook hook : pluginHooks) {
-                        instance.hookRunner.addHook(hook);
-                    }
-                } catch (Exception e) {
-                    logger.error(PLUGIN_GET_HOOKS_ERROR, pluginName(plugin));
-                }
-            }
-
-            List<RegistrationCompleteResult.Failure.PluginFailure> pluginFailures = new ArrayList<>();
-            for (Plugin plugin : instance.plugins) {
-                try {
-                    plugin.register(instance, metadata);
-                } catch (Exception e) {
-                    pluginFailures.add(new RegistrationCompleteResult.Failure.PluginFailure(pluginName(plugin), e.getMessage(), e));
-                    logger.error(PLUGIN_REGISTER_ERROR, pluginName(plugin));
-                }
-            }
+            List<RegistrationCompleteResult.Failure.PluginFailure> pluginFailures =
+                    instance.registerPlugins(instance.plugins, metadata);
 
             RegistrationCompleteResult pluginsRegistrationResult = pluginFailures.isEmpty()
                     ? RegistrationCompleteResult.success()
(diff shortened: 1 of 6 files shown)


launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/LDClient.java
```diff
@@ -895,4 +896,57 @@ static LDLogger getSharedLogger() {
...
+            logger.error("Exception thrown registering plugin " + pluginName(plugin) + ".");
+            result = RegistrationCompleteResult.failure(Collections.singletonList(
+                    new RegistrationCompleteResult.Failure.PluginFailure(pluginName(plugin), e.getMessage(), e)));
+        }
+
+        // The hooks go live only once register has succeeded, so a plugin whose registration failed
+        // never contributes hooks, and a plugin's own hooks do not observe its register call.
+        if (result instanceof RegistrationCompleteResult.Success) {
+            for (Hook hook : pluginHooks) {
+                hookRunner.addHook(hook);
```
tanderson-ld: Why are hooks added after register here?  Is that a different order than in the init flow?

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/LDClient.java
```diff
@@ -895,4 +896,57 @@ static LDLogger getSharedLogger() {
...
+            throw new NullPointerException("plugin must not be null");
+        }
+
+        EnvironmentMetadata metadata = environmentMetadata;
+
+        List<Hook> pluginHooks;
+        try {
+            pluginHooks = plugin.getHooks(metadata);
+        } catch (Exception e) {
+            logger.error("Exception thrown getting hooks for plugin " + pluginName(plugin) + ". Unable to get hooks, plugin will not be registered.");
```
tanderson-ld: Consider making constants for logging strings to ensure both the init registration flow and here are using the same strings and can't diverge by accident.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/LDClient.java
tanderson-ld, line 1: Is it ok that the order/params of the various plugin methods is different in these two scenarios?

Config-time, plugins = [A, B, C]:

A.getHooks → install; B.getHooks → install; C.getHooks → install
A.register; B.register; C.register
result = aggregate(A,B,C)
A.onPluginsReady(result); B.onPluginsReady(result); C.onPluginsReady(result)

Three sequential registerPlugin calls:

A.getHooks (held); A.register; install A's hooks; A.onPluginsReady(result_A)
B.getHooks (held); B.register; install B's hooks; B.onPluginsReady(result_B)
C.getHooks ...

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/LDClient.java
```diff
@@ -895,4 +896,57 @@ static LDLogger getSharedLogger() {
...
+        // The hooks go live only once register has succeeded, so a plugin whose registration failed
+        // never contributes hooks, and a plugin's own hooks do not observe its register call.
+        if (result instanceof RegistrationCompleteResult.Success) {
+            for (Hook hook : pluginHooks) {
+                hookRunner.addHook(hook);
+            }
+        }
+
+        try {
+            plugin.onPluginsReady(result, metadata);
```
tanderson-ld: Discussed, we're ok with removing this call since it appears to be initialization oriented and this register function is not at init time.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/LDClient.java
tanderson-ld: Does the adding of hooks have to be transactional or lock protected?  Double check to see if there is any existing risk in the init flow whereby something could lead to a partial set of hooks getting invoked.

tanderson-ld: Throwing register leaves hooks installed in the init logic, but that is different here.

tanderson-ld: I think HookRunner needs a bulk `addHooks` method to make sure these are transactional.  Otherwise a plugin with multiple hooks that all need to be registered together could face a race condition where Hook1 is registered, but Hook2 is not and a LDClient.variation... call is made in the interrim.

May also make sense to use this addHooks method in the init flow too instead of doing what it is doing today with the loop over the hooks with individual add calls.

tanderson-ld: Double check interaction with the hook runner is very very fast, evaluation calls will also need to be interacting with this list and so there will be impact of calling add hooks at runtime.

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/LDClient.java
abelonogov-ld: Yes, it is completely ok, considering onPluginsReady depreciation

launchdarkly-android-client-sdk/src/main/java/com/launchdarkly/sdk/android/integrations/Plugin.java
```diff
@@ -48,6 +48,18 @@ public List<Hook> getHooks(EnvironmentMetadata metadata) {
...
+    /**
+     * Called once every plugin configured on {@link com.launchdarkly.sdk.android.LDConfig} has been
+     * registered, reporting whether they all succeeded.
+     *
+     * @param result   the outcome of registering that batch of plugins
+     * @param metadata metadata about the environment where the plugin is running.
+     * @deprecated This reports on a batch of plugins registered together, so it has no meaning for
+     *             {@link LDClient#registerPlugin(Plugin)}, which registers a single plugin and does
+     *             not call it. Do work that needs the client in {@link #register} instead, which
+     *             both paths call.
```
tanderson-ld: This deprecation reason should not talk about specific usages of the abstract class.

You can just say "onPluginsReady was necessary in the past, but has since been determined to be problematic and is being removed in the next major version."

tanderson-ld (approved):

